PR #9, opened 2015-02-20
Update the list of warnings for MSVC builds.

sas:
This commit removes some /wd arguments that were useless (warnings
enabled only with /W4, but we build with /W3), and disables some type
casting warnings.


sas:
@as106 reviewed this.

PR #10, opened 2015-02-24
Android builds.

sas:
Changes for Android builds, along with build script, cmake toolchain file, and travis configuration for automated builds.


strager:
bon bon bon

lgtm

PR #12, opened 2015-02-25
Use ptsname_r instead of ptsname.

sas:
Using ptsname gets us a warning on Android builds because it is not
reentrant.


strager:
lgtm

PR #15, opened 2015-02-27
Fix a test failure on Linux-x86_64

leofanyp:
ds2 needs to link with a thread library to run correctly
with llgs test.


sas:
LGTM, will merge.

PR #18, opened 2015-03-04
Add release builds to travis.

sas:
Some warnings appear only when building release (rare), and some bugs
only occur with release builds (enabled by some optimizations). This
change makes everything build both release and debug so we can catch
these.


strager:
lgtm

PR #19, opened 2015-03-07
ARM64 support.

sas:


strager:
lgtm

Definitions/arm64.json
```diff
@@ -0,0 +1,776 @@
...
+                "format"                : "hex",
+
+                "base-gdb-reg-number"   : 0,
+                "base-gcc-reg-number"   : 0,
+                "base-dwarf-reg-number" : 0,
+
+                "referencing-sets"      : [ "pseudo-gpr32" ]
+            },
+
+            "r0"  : {
```
chancila: why r instead of x with r aliases?

strager:
It seems an AArch64 ds2 will not be able to debug an ARM program. Is that correct and intended?

strager:
lgtm, but the gdb special case is confusing to me.

strager:
(Sorry for the delay in reviewing.)

sas:
Yes.

On Wed, Apr 1, 2015 at 5:49 PM, strager <notifications@github.com> wrote:

> It seems an AArch64 ds2 will not be able to debug an ARM program. Is that
> correct and intended?
>
> —
> <https://github.com/facebook/ds2/pull/19#issuecomment-88676283>.
>



-- 
Stephane Sezer

sas:
As far as I can see there's been only two comments on this branch: x vs r for register names, which sounds pretty arbitrary, and the gdb special case thing. I'll take a look at the gdb code path a bit later. For now it's just mimicking whatever we have for arm already and I'd rather checkin this branch before it goes stale.

PR #21, opened 2015-04-13
Update patent grant.

sas:


sas:
CC @caabernathy @jamesgpearce

PR #23, opened 2015-04-14
Lower checksum failure logs from Warning to Info.

sas:
LLGS test suite always sends packets with 0 checksum. We end up printing
a ton of useless warnings in this case.


strager:
wowz

Is there an llgs bug open for this?

lgtm

sas:
It's not a bug, it's some sort of expected behavior. I did fix some of
these checksums, but given that no one really cares, it's pretty much a
waste of time.

On Tue, Apr 14, 2015 at 2:21 PM, strager <notifications@github.com> wrote:

> wowz
>
> Is there an llgs bug open for this?
>
> lgtm
>
> —
> <https://github.com/facebook/ds2/pull/23#issuecomment-93066715>.
>



-- 
Stephane Sezer

PR #35, opened 2015-06-23
Rework the way stop codes are handled.

sas:


strager:
lgtm through eae340920a08d16cc3cdbeb30b6a74854416e6d0

PR #44, opened 2015-07-22
Add stubs for qModuleInfo.

sas:
I thought qModuleInfo was going to be required for Windows support, but we can probably do away without it. Still adding the current work I have so we have the stubs if we wanna implement it later on.


strager:
lgtm

PR #45, opened 2015-07-23
Handle Thread Exit Cleanly

BeyondSora:
This change ensures ds2 handles thread exit cleanly.

Previously DS2 updates a thread and then checks for `_stopInfo.event`. This leads to a problem where a thread placeholder remains for too long and when `updateState()` invokes `ReadStat`, there is no such thread to be found in `/proc/...`. This change simply checks `_stopInfo.event` first and return early if the thread has already exited.


sas:
Make sure the body of your commit message is at 72 columns. Otherwise, this patch looks good.

BeyondSora:
Updated per comments.

sas:
Merged this commit. Thanks for the contribution and sorry for the delay :)

BeyondSora:
Awesome. Thank you!!

PR #46, opened 2015-07-23
On Linux don't terminate process when detached

### LINKED ISSUE #43: Detaching from inferior seems to kill it.
When we detach from a debugged process, the process dies. This is probably due to something wrong we're doing during the `PTRACE_DETACH`. Need to investigate.

Woody17:
Fixes #43,
When the Linux process (ds2/Sources/Target/Linux/Process.cpp) deallocates it calls terminate on the process even when ds2 doesn't own the process.

- Removes the call to terminate() in the destructor to fix this


sas:
Committed as 73884a345bda2cc4479ee5995872466783a6cd76. Thanks for the fix. This also resolves issue#43.

PR #52, opened 2015-09-22
Build regsgen2 on travis.

sas:


strager:
Fix security problems. Otherwise, lgtm.

sas:
llvm.org is down, which is why the "Style" build fails.

chancila:
wow so great +2

PR #66, opened 2015-10-24
FreeBSD support

jceel:
Debugging single- and multi-threaded programs work, but the latter is likely sub-optimal in terms of performance. Allocating memory inside debugged process doesn't work (requires different assembly code). POSIX stringify implementation should be probably split into separate implementations for Linux and FreeBSD.


Headers/DebugServer2/Target/Process.h
```diff
@@ -17,6 +17,8 @@
 #include "DebugServer2/Target/Linux/Process.h"
 #elif defined(OS_WIN32)
 #include "DebugServer2/Target/Windows/Process.h"
+#elif defined(__FreeBSD__)
```
sas: You should use OS_FREEBSD here.

Headers/DebugServer2/Target/Process.h
```diff
@@ -28,6 +30,8 @@ namespace Target {
 using Linux::Process;
 #elif defined(OS_WIN32)
 using Windows::Process;
+#elif defined(__FreeBSD__)
```
sas: Ditto.

Headers/DebugServer2/Target/ProcessDecl.h
```diff
@@ -28,6 +28,9 @@ __FORWARD_DECLARE(Linux, Thread);
 #elif defined(OS_WIN32)
 __FORWARD_DECLARE(Windows, Process);
 __FORWARD_DECLARE(Windows, Thread);
+#elif defined(__FreeBSD__)
```
sas: Ditto.

Headers/DebugServer2/Target/Thread.h
```diff
@@ -17,6 +17,8 @@
 #include "DebugServer2/Target/Linux/Thread.h"
 #elif defined(OS_WIN32)
 #include "DebugServer2/Target/Windows/Thread.h"
+#elif defined(__FreeBSD__)
```
sas: And so on... :)

Sources/GDBRemote/Session.cpp
```diff
@@ -24,6 +24,8 @@
 
 #if defined(OS_LINUX)
 #define UNPACK_ID(STR) std::strtoul(STR, nullptr, 10)
+#elif defined(__FreeBSD__)
+#define UNPACK_ID(STR) std::strtoul(STR, nullptr, 10)
```
sas: Any reason why this (and FORMAT_ID below) is not merged with the Linux one?

Headers/DebugServer2/Target/Process.h
jceel: Yeah, right, I've changed that in some places but apparently not all of them :)

Sources/GDBRemote/Session.cpp
jceel: No, I don't think there's any reason. I can't remember now why I did it that way. Anyhow, maybe what we need is OS_POSIX in places like that?

sas:
This looks very good. Apart from the small nits I posted earlier (which can be fixed later), I think this is good to go in.

I squashed the commits, ran a pass of clang-format, and pushed to master as 68973a5424a2337fe6bdfa8433e879d2e489b8cd and 8f87591e92afa79f8c6e795b1de5a5202c4f0079.

Thanks a lot for this contribution!

PR #72, opened 2015-11-19
Logging improvements.

sas:
Main improvement here is to be able to log to android logcat.

### DIFF
--- a/Sources/Utils/Log.cpp
+++ b/Sources/Utils/Log.cpp
@@ -19,6 +19,7 @@
 #include "DebugServer2/Host/FreeBSD/ExtraWrappers.h"
 #endif
 
+#include <dlfcn.h>
 #include <sstream>
 
 namespace {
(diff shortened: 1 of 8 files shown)


Sources/Utils/Log.cpp
```diff
@@ -33,6 +35,33 @@ FILE *sOutputStream = stderr;
 #endif
 }
 
+#if defined(__ANDROID__)
+#include <android/log.h>
+
+static bool androidLogcat(int level, char const *functag, char const *message) {
+  static const std::map<int, android_LogPriority> levelConverter = {
+      {ds2::kLogLevelPacket, ANDROID_LOG_VERBOSE},
```
chancila: PQN switch?

strager:
lgtm! c'bon.

I like using dlopen for this, and I like how C++ made it (relatively) pretty to use.

PR #76, opened 2015-11-24
Implement jThreadsInfo packet

fjricci:
Allows for batch-wise communication of thread stop information to lldb


fjricci:
So other packets also do this:
Handle_qThreadStopInfo
Handle_vAttachOrWait
Handle_vAttachWait

fjricci:
I have no problem changing it, but it seemed to be the convention

sas:
If you rebase this on top of master, you should be able to pass all the tests.

PR #77, opened 2015-11-24
Define a dummy __srget() to work around an Android NDK issue.

sas:
See comments in the source for more details on what issue we're trying
to fix.


strager:
wow

+2

PR #79, opened 2015-11-25
Move to container-based travis infra.

sas:


sas:
This doesn't look like it's ready for us right now. Will revisit later.

PR #80, opened 2015-11-25
Get newer cmake from cmake.org.

sas:


strager:
bon, lgtm

Maybe check the md5?

strager:
Would be nice to describe why you need a newer CMake.

sas:
Meh.

On Tue, Nov 24, 2015 at 9:32 PM, strager <notifications@github.com> wrote:

> bon, lgtm
>
> Maybe check the md5?
>
> —
> <https://github.com/facebook/ds2/pull/80#issuecomment-159495887>.
>



-- 
Stephane Sezer

PR #86, opened 2015-12-02
Fix memory read error in EnumerateLinkMap

fjricci:
Currently, the PATH_MAX read will always go out of bounds and fail,
and no shared libraries will be found. Instead of requiring a
successful read of PATH_MAX bytes, accept a successful read of
any number of bytes.

Fixes enumerateSharedLibraries.


sas:
Cool stuff, just a small nit for the string check. Also, can you squash your three patches together?

sas:
Cool. Gonna check this in now.

sas:
Committed as e51e3de81182ce7325ec2be35e3ffb678e130d9b.

PR #87, opened 2015-12-02
Get rid of uses of wait4.

sas:
We can use waitpid instead, which seems to be easier to find on some
sysroots, and provides the functionality of wait4 minus rusage (which we
never used).


fjricci:
looks good to me

PR #89, opened 2015-12-03
Stop sending success responses for qSymbol

fjricci:
DS2 doesn't support this packet, but was sending responses
indicating that it does.


sas:
lgtm. will merge.

PR #91, opened 2015-12-08
Use DS2_UNREACHABLE in log level conversion.

sas:
This avoids a warning about androidLevel being potentially
uninitialized.


as106:
lgtm

PR #92, opened 2015-12-08
[WIP] Switch travis to Ubuntu Trusty.

sas:
Trusty is finally available on Travis. Use it!


sas:
@fjricci is taking over this patch.

PR #93, opened 2015-12-09
Directly link against liblog on android.

sas:
The dlopen stuff is usless as the binaries generated with the android
toolchain won't run on other systems anyway.


sas:
Ah, I need to add ENABLE_LOGCAT to the X86 thing.

It's not really better, it's just that we need it that way if we want to link directly against logcat.

fjricci:
Looks good to me

PR #95, opened 2015-12-10
Always dump messages to logcat.

sas:
Even if the log level doesn't allow, dump debug info to logcat.


fjricci:
I like this. Looks good.

PR #97, opened 2015-12-11
Fix compilation with user_fpxregs_struct on android-x86

fjricci:
I think the aosp toolchain changed, because sys/user.h now defines this struct.


sas:
I'm fine with this patch, but the issue is that these were not defined in earlier versions of the android sysroot. I'm not sure if we should just push your change and drop support for older versions, or if we should try to do some checking and enable (or not) these local definitions.

sas:
Alright, we don't care about older version; I committed this as 189d49371caebfae180fe21bc42ed77d0c996468.

PR #98, opened 2015-12-11
Add a function to stringify error codes.

sas:


fjricci:
Nvm, now I see it matches the other stringify classes.

Looks good. And makes life easier.

sas:
Thanks!

PR #99, opened 2015-12-11
Prevent android logging from overflowing logcat buffer

fjricci:
From android/log.h:

"Log message text may be truncated to less than an implementation-specific
limit (e.g. 1023 characters max)."

Since logcat appends newline characters to the buffer anyway,
split the message string on newlines to prevent awkwardly-split string.


sas:
Committed as c0cf6d610b0633647b763013787b115a50e6713c.

PR #100, opened 2015-12-12
Make necessary fixes and performance optimizations for SVR4 packet

fjricci:


Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -203,10 +201,11 @@ EnumerateLinkMap(ELFProcess *process, Address addressToDPtr,
     if (error != ds2::kSuccess)
       return error;
 
+    shlib.main = (process->loadBase() == linkMap.baseAddress);
```
vreis: This is not what I had in mind when we talked about it. linkMap.baseAddress is the relocation applied to the executable, and it will be zero for non-PIE executables. loadBase seems to be the lowest virtual address that is mapped in the process. Are you sure this is working for non-PIE executables?

vreis: Discussed this offline with Francis: this will work for PIE executables, and I think that's better than the current solution so I'm fine accepting this diff. And doing the right thing here probably requires changing lldb.

sas:
We don't need this anymore, detection of main executable is fixed.

PR #101, opened 2015-12-12
Fix some nasty issue in ds2.

sas:
Not sure how to describe this better :/


strager:
Patches lgtm except c35d7dd55800c80cc1e2f8005fd9acd63680f75d. I don't see `Stringify::Ptrace` used anywhere.

sas:
I'll commit this without the last patch that disables jThreadsInfo, as we found the bug in lldb and fixed it.

PR #102, opened 2015-12-17
LLDB 3.7 will fail to locate libraries if arch isn't exact match

fjricci:


sas:
I'm pretty sure lldb is being dumb here, but this is a good fix nonetheless. Will merge.

sas:
Committed as cca10a45b4fe72582dccef890a5aa6b9d47dc948.

PR #106, opened 2015-12-22
Enable interrupt packet handling on windows

fjricci:
Allows ds2 to kill processes that are not already stopped.


sas:
+2. Will merge in a few days.

sas:
Not sure why travis doesn't run on this :/

PR #107, opened 2015-12-23
Implement attaching to running process on Windows

fjricci:


sas:
Pushed a slightly different version of this. See PR#117.

PR #111, opened 2015-12-30
Always build all the linux syscall wrappers as weak.

sas:
This help prevents issues with ds2 being built with a certain sysroot
(that has some functions) and then run on a device with an older OS
versions that doesn't have these symbols.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ set(HOST_POSIX_SOURCES
 
 set(HOST_Linux_SOURCES
     ${HOST_POSIX_SOURCES}
+    Sources/Host/Linux/ExtraWrappers.cpp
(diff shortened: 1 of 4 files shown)


strager:
lgtm, but I'm skeptical about the `weak`ness.

sas:
No these really do work :)

I'll move the __attribute__ bits to `CompilerSupport.h` before checking
this in though.

On Tue, Dec 29, 2015 at 5:13 PM, strager <notifications@github.com> wrote:

> lgtm, but I'm skeptical about the weakness.
>
> —
> <https://github.com/facebook/ds2/pull/111#issuecomment-167913594>.
>



-- 
Stephane Sezer

strager:
lgtm

PR #114, opened 2016-01-04
Remove an unused parameter from stop info query function

fjricci:


sas:
Nice, what was this list supposed to mean? I'll merge this.

fjricci:
It was a remnant from an old implementation of jThreadsInfo, where it meant "list" of threads vs 1 thread. But I ended up changing the implementation and must have forgotten to remove the parameter.

PR #119, opened 2016-01-12
Update travis to use lldb-server tests from lldb 3.7

fjricci:


fjricci:
This will probably conflict with #120, let's get that one in first and then update this one if necessary.

fjricci:
Migrated to #122

PR #123, opened 2016-01-19
[WIP] Docker

fjricci:


fjricci:
Docker doesn't use trusty :(

PR #130, opened 2016-01-21
Enable full lldb test suite on travis

fjricci:


sas:
LGTM.

fjricci:
dosep.py seems like the better option - #132

PR #136, opened 2016-01-25
Implement IO redirection packets

fjricci:


Sources/GDBRemote/DebugSessionImpl.cpp
```diff
@@ -572,6 +572,29 @@ ErrorCode DebugSessionImpl::onXferRead(Session &session,
...
+                                         std::string const &path) {
+  bool success = false;
+  switch (fileno) {
+  case 0:
+    success = _spawner.redirectInputToFile(path);
+    break;
+  case 1:
+    success = _spawner.redirectOutputToFile(path);
+    break;
+  case 2:
```
kastiglione: Can `STD{IN,OUT,ERR}_FILENO` constants be used?

fjricci: For some reason, the ProcessSpawner methods seem to use hardcoded 0,1,2. But using constants seems reasonable.

sas:
Committed as 243fd795229d41068e39031c78a2eee7162d99ae..bc8a70abe9cf328e170c760245e6f0ee7605de67.

PR #150, opened 2016-01-28
Include jstopinfo in qThreadStopInfo

a20012251:
Add the jstopinfo field to the qThreadStopInfo response.

It's a json object with the hex tid of each thread along with its stop reason. Handle_jThreadsInfo also fetches the information of all threads, so I refactored this logic out into a new function. It's appended as
 
```
jstopinfo:[{... json object...}];
```

I also refactored the encoding method for StopCode. There's a new method encodeWithJStopInfo that appends the jstopinfo to the stream generated by the generic encoder. I'm tempted to move the encoding logic to a new class, but I don't know if that makes a lot of sense.

I tested it by stopping a multi-threaded dummy program with a breakpoint. The log of the packets related to qThreadStopInfo showed:

```
<  23> send packet: $qThreadStopInfo13e4#f8
< 561> read packet: $T00thread:13e4;name:a.out;core:0;threads:13e4,13f1;00:00feffffffffffff;01..lots of numbers...;;17:00000000;jstopinfo:[{"tid":"13e4","reason":"none"}],{"tid":"13f1","reason":"breakpoint"}]];#f8

```
where the suffix contains the jstopinfo.

### DIFF
--- a/Headers/DebugServer2/GDBRemote/DebugSessionImpl.h
+++ b/Headers/DebugServer2/GDBRemote/DebugSessionImpl.h
@@ -155,6 +155,12 @@ protected:
   ErrorCode queryStopCode(Session &session, ProcessThreadId const &ptid,
                           StopCode &stop) const;
 
+protected:
+  ErrorCode fetchStopInfoForAllThreads(Session &session,
+                                       std::vector<StopCode> &stops,
+                                       StopCode &processStop);
+  ErrorCode createThreadsStopInfo(Session &session, JSArray &threadsStopInfo);
+
 private:
   ErrorCode spawnProcess(StringCollection const &args,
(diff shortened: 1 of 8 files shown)


a20012251:
just updated the code, now i'll test with osmeta's lldb

fjricci:
Looks good to me, once you check it against lldb.

a20012251:
Now jstopinfo is sent hexified, just as lldb expects

a20012251:
i've fixed the hex encoding part, it's simpler now

a20012251:
now there's a single commit

sas:
Committed as 8c0555723746eda7cb08a979b945bcbcab266cfa.

PR #155, opened 2016-01-29
Add verbose logging to dosep.py

fjricci:
Should prevent most test flakiness on dosep


fjricci:
Noticed odd behavior looking at the passing vs failing test outputs. It looks like the last test to show up as having run is "TestSTTYBeforeAndAfter.CommandLineCompletionTestCase". If this test is run last, the suite passes, if it's run in the middle, the suite fails. Seems very strange to me.

fjricci:
Yeah, it's a newline, because otherwise we write on the same line as "x out of y test suites processed"

PR #160, opened 2016-02-03
Add MinGW builds to travis.

sas:
We don't ship MinGW binaries for Windows, but this offers some coverage
for Windows builds as Travis doesn't support Windows + VS.


sas:
This should be merged after PR#161 goes in. There's a bunch of warnings with MinGW builds.

fjricci:
Committed as ebaef4fe3bc1ba47aa552677f6be1c8e964373b4

PR #163, opened 2016-02-05
Use IP for localhost.

sas:
This works around issues with name resolution.


kastiglione:
What about IPv6? :trollface:

fjricci:
I don't think we support IPv6 anyway. :laughing: Although this just affects the default host used if no host is specified, so if the user really wants to use IPv6, they could specify it. (The reason we did this is that name resolution was failing for localhost on Tizen...)

PR #164, opened 2016-02-05
Enable ipv6 support.

sas:


Sources/Host/Common/Socket.cpp
```diff
@@ -46,15 +45,26 @@ Socket::Socket(SOCKET handle)
...
     _lastError = SOCK_ERRNO;
   }
 
+  // On most Linux systems, IPV6_V6ONLY is off by default, but on some systems
+  // it's on, so turn it off to be able to receive both IPv6 and IPv4
+  // connections when we listen on IN6ADDR_ANY.
+  if (af == AF_INET6) {
+    int no = 0;
+    int res = setsockopt(_handle, IPPROTO_IPV6, IPV6_V6ONLY, &no, sizeof(no));
+    if (res != 0) {
```
a20012251: It's generally discouraged to use temp variables unless it's necessary, you can just put the setsockopt call inside the if

kastiglione:
looks good

PR #172, opened 2016-02-09
onQuery methods should be const, make it happen

fjricci:


Headers/DebugServer2/GDBRemote/Types.h
```diff
@@ -208,6 +208,30 @@ struct ProgramResult {
...
+  inline void clear() {
+    index = 0;
+    ptids.clear();
+  }
+
+  inline void addPid(ProcessId pid) { ptids.push_back(ProcessThreadId(pid)); }
+
+  inline void reset(std::vector<ThreadId> const &tids) {
+    clear();
+    for (auto &tid : tids)
```
a20012251: for uniformity I think we should always use { } 
you can also use emplace_back in this case, avoiding some copies

a20012251:
I just have a weird feeling about using the struct's members' methods instead of encapsulating them in new struct methods, but that might just be a habit from other languages

fjricci:
If you look at the other structs in the file and how they're used, this is pretty standard behavior in our codebase (for better or for worse)

fjricci:
Committed as 147b6e6913d88f18a19ae7e10e1b8e495a595836...7f1de1c52e7d731c9a71212b49090529fafadd6d

PR #175, opened 2016-02-10
Use g++ 5 on Travis

fjricci:


sas:
This is odd, the clang builds are failing because of the `override` keyword missing. These should be warnings but not errors. What happened?

fjricci:
I think you may have accidentally mixed up the builds. The clang ones are passing and the gcc-5 ones are failing, which is the expected behavior. I'll rebase to fix the gcc builds

sas:
Aha! Possible. I see you rebased the patch also. Feel free to submit when travis is green.

PR #176, opened 2016-02-10
Dump all ptrace commands to debug log.

sas:
This is verbose but really helpful for debugging.


sas:
Looks like this passes tests now.

PR #177, opened 2016-02-12
Use getpeer name on client sockets.

sas:
This is important so we display the right information during
reverse-connect.

### DIFF
--- a/Sources/Host/Common/Socket.cpp
+++ b/Sources/Host/Common/Socket.cpp
@@ -20,19 +20,19 @@
 #include <ws2tcpip.h>
 #define SOCK_ERRNO WSAGetLastError()
 #define SOCK_WOULDBLOCK WSAEWOULDBLOCK
-#else
+#elif defined(OS_LINUX) || defined(OS_FREEBSD)
 #include <arpa/inet.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <netdb.h>
 #include <netinet/in.h>
 #include <poll.h>
-#include <sys/socket.h>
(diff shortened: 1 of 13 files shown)


Sources/Host/Common/Socket.cpp
```diff
@@ -355,10 +355,10 @@ std::string Socket::port() const {
     return std::string();
   }
 
+  auto func = listening() ? ::getsockname : ::getpeername;
```
a20012251: I'd create a method that returns the function, just follow the DRY rule ~~~

Sources/Host/Common/Socket.cpp
```diff
@@ -355,10 +355,10 @@ std::string Socket::port() const {
...
   struct sockaddr_storage ss;
   socklen_t sslen = sizeof(ss);
-  if (::getsockname(_handle, reinterpret_cast<struct sockaddr *>(&ss), &sslen) <
-      0) {
+  if (func(_handle, reinterpret_cast<struct sockaddr *>(&ss), &sslen) < 0) {
```
a20012251: there's code duplication
I'd move lines 358, 360 and 361 to a new function, same as above, to follow the DRY rule

sas:
Committed the first commits (the ones that passed travis) as 22144148e80d2a572a01da5bca423b337f1f836e..9f3f60b0889fc0c6c0a4fd66b81051d601db9897.

sas:
Committed the last patch as facdb90caf613b18da4bf51b73d05487a27d518e.

PR #182, opened 2016-02-19
Fix usage of uninitialized value

kokarez:
`si` was used because the code comme from the linux version. But in the
freebsd one, we don't get siginfo_t directly, we get it from an other
structure. So here all usage of si are non correct.

Please not that I didn't compile this change since the build sound to be broken for Freebsd. I've root the build issue to the commit [a5bcf6f]

[a5bcf6f] https://github.com/facebook/ds2/commit/a5bcf6f62704adf4f03c012605e5f84ad8c9e820


sas:
@facebook-github-bot shipit

sas:
Committed as 2f2a000c384dc9feed8296da07f5e418b1be320c.

PR #183, opened 2016-02-19
Fix FreeBSD build

kokarez:
2 fixes:
 - Disable linux specific feature in elf's code https://github.com/facebook/ds2/commit/a5bcf6f62704adf4f03c012605e5f84ad8c9e820
    * For this fix I try to stay as close as possible to the code before the faulty commit, the only difference is we will set back the _hack_ after getting the MemoryInfo.
 - Remove hang usage in all the call to Process::wait https://github.com/facebook/ds2/commit/02aa8b1970bae5ae58d50c0f4a719a51d132b328

test lldb on MacOSX, ds2 on freebsd
```bash
bash-3.2$ lldb a.out 
(lldb) target create "a.out"
Current executable set to 'a.out' (x86_64).
(lldb) gdb-remote xx.xx.xx.159:4242
Process 98820 stopped
* thread #1: tid = 0x1884d, 0x0000000800602420, name = 'a.out', stop reason = signal SIGTRAP
    frame #0: 0x0000000800602420
->  0x800602420: xorq   %rbp, %rbp
    0x800602423: subq   $0x18, %rsp
    0x800602427: movq   %rdi, %r12
    0x80060242a: movq   %rsp, %rsi
(lldb) c
Process 98820 resuming
Process 98820 exited with status = 0 (0x00000000) 
(lldb)  
```

### DIFF
--- a/Sources/Target/FreeBSD/Process.cpp
+++ b/Sources/Target/FreeBSD/Process.cpp
@@ -48,7 +48,7 @@ ErrorCode Process::initialize(ProcessId pid, uint32_t flags) {
   // Wait the main thread.
   //
   int status;
-  ErrorCode error = ptrace().wait(pid, true, &status);
+  ErrorCode error = ptrace().wait(pid, &status);
   DS2LOG(Debug, "wait result=%d", error);
   if (error != kSuccess)
     return error;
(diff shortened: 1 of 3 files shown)


sas:
This is good. Can you make just two separate commits (on the same PR is fine), one for the first fix (hang argument), one for the second fix (MRI stuff)?

sas:
@facebook-github-bot shipit

sas:
LGTM except for the added comment that seems useless.

sas:
Committed as 2c1f9fbf9fdcd9b9d066db1c96e2c1df8fa2907a..e4c8fb61ff83216e532067df686e752131e377ae.

PR #184, opened 2016-02-19
Remove redundant gcc-5 linux toolchain file

fjricci:


sas:
@facebook-github-bot shipit

PR #186, opened 2016-02-20
Fix stringification of ptrace regs commands on centos

fjricci:


sas:
Can you just add a comment under the `#if defined(OS_LINUX)` part that explains why we need this?
Otherwise, LGTM.

sas:
@facebook-github-bot shipit

sas:
Committed as 4874b8ec1e17474aece8e33a0f5b077a51505e89.

PR #197, opened 2016-02-20
Last bits of Mac OS X support

kokarez:
Hi, first of all, sorry it's not a lightweight patch :) Ask me how you prefer it and I will do it!

MacOSX support status:
- Compile
- Read most of the process info with LibProc
- PTrace is fully (?) support (since the 10% offered by mac was already here)
- Mach base communication done task/thread
- Can read the register of a thread
- Can read in the memory of the application
- Certification work (needed to use mach arch)

Todo:
- Compile on travis (travis already propose OSX env)
- write in memory
- set cpu state
- get memory region
- clean the certification plist.
- auto test :)

Here's an example with a basic app:

``` gdb
(lldb) target create "./a.out"
Current executable set to './a.out' (x86_64).
(lldb) gdb-remote localhost:4243
Process 46277 stopped
* thread #1: tid = 0xb4c5, 0x00007fff668c8000, stop reason = signal SIGTRAP
    frame #0: 0x00007fff668c8000
->  0x7fff668c8000: popq   %rdi
    0x7fff668c8001: pushq  $0x0
    0x7fff668c8003: movq   %rsp, %rbp
    0x7fff668c8006: andq   $-0x10, %rsp
(lldb) x/20i 0x7fff668c8000
->  0x7fff668c8000: 5f                    popq   %rdi
    0x7fff668c8001: 6a 00                 pushq  $0x0
    0x7fff668c8003: 48 89 e5              movq   %rsp, %rbp
    0x7fff668c8006: 48 83 e4 f0           andq   $-0x10, %rsp
    0x7fff668c800a: 48 83 ec 10           subq   $0x10, %rsp
    0x7fff668c800e: 8b 75 08              movl   0x8(%rbp), %esi
    0x7fff668c8011: 48 8d 55 10           leaq   0x10(%rbp), %rdx
    0x7fff668c8015: 4c 8b 05 ec 7a 03 00  movq   0x37aec(%rip), %r8
    0x7fff668c801c: 48 8d 0d dd ff ff ff  leaq   -0x23(%rip), %rcx
    0x7fff668c8023: 4c 29 c1              subq   %r8, %rcx
    0x7fff668c8026: 4c 8d 05 d3 ef ff ff  leaq   -0x102d(%rip), %r8
    0x7fff668c802d: 4c 8d 4d f8           leaq   -0x8(%rbp), %r9
    0x7fff668c8031: e8 40 00 00 00        callq  0x7fff668c8076
    0x7fff668c8036: 48 8b 7d f8           movq   -0x8(%rbp), %rdi
    0x7fff668c803a: 48 83 ff 00           cmpq   $0x0, %rdi
    0x7fff668c803e: 75 10                 jne    0x7fff668c8050
    0x7fff668c8040: 48 89 ec              movq   %rbp, %rsp
    0x7fff668c8043: 48 83 c4 08           addq   $0x8, %rsp
    0x7fff668c8047: 48 c7 c5 00 00 00 00  movq   $0x0, %rbp
    0x7fff668c804e: ff e0                 jmpq   *%rax
```

It's look like code, and after multiple reboot, I get the same code. So I suppose it's working :D 

Thanks in advance for your time, and have fun reading!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,17 +57,17 @@ matrix:
       env: TARGET="Linux-X86_64"   LLDB_TESTS="all"  CLANG="1" RELEASE="1"
     # Darwin-X86_64 targets
     - os: osx
-      osx_image: xcode7
+      osx_image: xcode7.3
+      env: TARGET="Darwin-X86_64" LLDB_TESTS="all" CLANG="1"
+    - os: osx
+      osx_image: xcode7.3
       env: TARGET="Darwin-X86_64" CLANG="0"
     - os: osx
       osx_image: xcode7
       env: TARGET="Darwin-X86_64" CLANG="1"
     - os: osx
-      osx_image: xcode7.3
+      osx_image: xcode7
       env: TARGET="Darwin-X86_64" CLANG="0"
-    - os: osx
-      osx_image: xcode7.3
-      env: TARGET="Darwin-X86_64" CLANG="1"
 
 before_install: ./Support/Testing/Travis/before-install.py
 install:        ./Support/Testing/Travis/install.py
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,16 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     set(ARCH_NAME "X86")
   endif ()
 endif ()
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
+  set(GENERATED_MACH_SOURCES
+      ${CMAKE_CURRENT_BINARY_DIR}/mach_excServer.c
+      ${CMAKE_CURRENT_BINARY_DIR}/mach_excUser.c
+  )
+  add_custom_command(OUTPUT ${GENERATED_MACH_SOURCES}
+    COMMAND mig ${CMAKE_SOURCE_DIR}/Sources/Host/Darwin/machGen.defs
+    DEPENDS ${CMAKE_SOURCE_DIR}/Sources/Host/Darwin/machGen.defs
+  )
+endif ()
 
 set(ARCHITECTURE_COMMON_SOURCES
     Sources/Architecture/RegisterLayout.cpp
@@ -116,6 +126,7 @@ set(HOST_Linux_SOURCES
     )
 
 set(HOST_Darwin_SOURCES
+    ${GENERATED_MACH_SOURCES}
     ${HOST_POSIX_SOURCES}
     Sources/Host/Darwin/LibProc.cpp
     Sources/Host/Darwin/Platform.cpp
--- a/Headers/DebugServer2/Host/Darwin/Mach.h
+++ b/Headers/DebugServer2/Host/Darwin/Mach.h
@@ -22,9 +22,25 @@ namespace ds2 {
 namespace Host {
 namespace Darwin {
 
+struct MachExcStatus {
+  uint32_t type;
+  uint32_t subtype;
+  uint32_t data;
+};
+
 class Mach {
+
+  struct MachMsg {
+    mach_msg_header_t hdr;
+    char payload[1024];
+  };
+
 public:
   virtual ~Mach() = default;
+
+protected:
+  mach_port_t _exc_port;
+  MachMsg _reply;
 
 public:
   ErrorCode readMemory(ProcessThreadId const &ptid, Address const &address,
@@ -57,6 +73,12 @@ public:
   ErrorCode getThreadInfo(ProcessThreadId const &tid, thread_basic_info_t info);
   ErrorCode getThreadIdentifierInfo(ProcessThreadId const &tid,
                                     thread_identifier_info_data_t *threadID);
+
+public:
+  ErrorCode setupExceptionChannel(ProcessId pid);
+  ErrorCode readException(MachExcStatus &status, bool timeout = true,
+                          thread_t *thread = nullptr);
+  bool exceptionIsFromThread(ProcessThreadId const &ptid);
 
 private:
   task_t getMachTask(ProcessId pid);
--- a/Headers/DebugServer2/Host/Darwin/Mach.h
+++ b/Headers/DebugServer2/Host/Darwin/Mach.h
@@ -36,9 +36,12 @@ class Mach {
   };
 
 public:
+  Mach(pid_t pid = -1);
   virtual ~Mach() = default;
 
 protected:
+  pid_t _pid;
+  task_t _task;
   mach_port_t _exc_port;
   MachMsg _reply;
 
@@ -65,8 +68,8 @@ public:
                    int signal = 0, Address const &address = Address());
 
 public:
-  ErrorCode getProcessDylbInfo(ProcessId pid, Address &address);
-  ErrorCode getProcessMemoryRegion(ProcessId pid, Address const &address,
+  ErrorCode getProcessDylbInfo(Address &address);
+  ErrorCode getProcessMemoryRegion(Address const &address,
                                    MemoryRegionInfo &info);
 
 public:
@@ -75,13 +78,13 @@ public:
                                     thread_identifier_info_data_t *threadID);
 
 public:
-  ErrorCode setupExceptionChannel(ProcessId pid);
+  ErrorCode setupExceptionChannel();
   ErrorCode readException(MachExcStatus &status, bool timeout = true,
                           thread_t *thread = nullptr);
   bool exceptionIsFromThread(ProcessThreadId const &ptid);
 
 private:
-  task_t getMachTask(ProcessId pid);
+  task_t getMachTask(ProcessId pid = -1);
   thread_t getMachThread(ProcessThreadId const &tid);
 };
 }
--- a/Headers/DebugServer2/Target/Darwin/MachOProcess.h
+++ b/Headers/DebugServer2/Target/Darwin/MachOProcess.h
@@ -39,7 +39,7 @@ public:
                                std::set<ds2::Target::Thread *>()) override;
 
 public:
-  Host::Darwin::Mach &mach();
+  Host::Darwin::Mach &mach(ProcessId pid = -1);
 
 protected:
   ErrorCode updateInfo() override;
--- a/Headers/DebugServer2/Target/Darwin/MachOProcess.h
+++ b/Headers/DebugServer2/Target/Darwin/MachOProcess.h
@@ -34,6 +34,11 @@ public:
       std::function<void(SharedLibraryInfo const &)> const &cb);
 
 public:
+  virtual ErrorCode resume(int signal = 0,
+                           std::set<ds2::Target::Thread *> const &excluded =
+                               std::set<ds2::Target::Thread *>()) override;
+
+public:
   Host::Darwin::Mach &mach();
 
 protected:
--- a/Headers/DebugServer2/Target/Darwin/Thread.h
+++ b/Headers/DebugServer2/Target/Darwin/Thread.h
@@ -40,11 +40,12 @@ protected:
   void updateState() override;
 
 public:
-  virtual ErrorCode step(int signal,
+  virtual ErrorCode suspend() override;
+  virtual ErrorCode step(int signal = 0,
                          Address const &address = Address()) override;
-
-public:
   virtual ErrorCode afterResume();
+  virtual ErrorCode resume(int signal = 0,
+                           Address const &address = Address()) override;
 
 public:
   virtual ErrorCode readCPUState(Architecture::CPUState &state) override;
--- a/Headers/DebugServer2/Target/Darwin/Thread.h
+++ b/Headers/DebugServer2/Target/Darwin/Thread.h
@@ -11,11 +11,11 @@
 #ifndef __DebugServer2_Target_Darwin_Thread_h
 #define __DebugServer2_Target_Darwin_Thread_h
 
+#include "DebugServer2/Host/Darwin/Mach.h"
 #include "DebugServer2/Target/POSIX/Thread.h"
 
 #include <csignal>
 #include <libproc.h>
-
 #include <mach/mach.h>
 #include <mach/mach_vm.h>
 #include <sys/types.h>
@@ -33,7 +33,10 @@ protected:
   Thread(Process *process, ThreadId tid);
 
 protected:
-  ErrorCode updateStopInfo(int waitStatus) override;
+  ErrorCode updateStopInfo(Host::Darwin::MachExcStatus &status);
+  ErrorCode updateStopInfo(int waitStatus) override {
+    return kErrorUnsupported;
+  };
   void updateState() override;
 
 public:
--- a/Sources/Host/Darwin/LibProc.cpp
+++ b/Sources/Host/Darwin/LibProc.cpp
@@ -59,7 +59,7 @@ std::string LibProc::GetThreadName(ProcessThreadId const &ptid) {
   static const std::string defaultName = "<unknown>";
   thread_identifier_info_data_t threadId;
   struct proc_threadinfo ti;
-  Mach mach;
+  Mach mach(ptid.pid);
 
   ErrorCode error = mach.getThreadIdentifierInfo(ptid, &threadId);
   if (error != kSuccess) {
--- a/Sources/Host/Darwin/Mach.cpp
+++ b/Sources/Host/Darwin/Mach.cpp
@@ -24,6 +24,42 @@
 #include <mach/thread_info.h>
 #include <sys/types.h>
 
+// Mach exception are manage by generated code, not a lib. Also the callback
+// to these system can't be register, it's hardlink. So we need to compile
+// these function to have access to the exception system. Also it's C...
+extern "C" {
+kern_return_t catch_mach_exception_raise_state(
+    mach_port_t exc_port, exception_type_t exc_type,
+    const mach_exception_data_t exc_data, mach_msg_type_number_t exc_data_count,
+    int *flavor, const thread_state_t old_state,
+    mach_msg_type_number_t old_stateCnt, thread_state_t new_state,
+    mach_msg_type_number_t *new_stateCnt) {
+  return KERN_FAILURE;
+}
+
+kern_return_t catch_mach_exception_raise_state_identity(
+    mach_port_t exc_port, mach_port_t thread_port, mach_port_t task_port,
+    exception_type_t exc_type, mach_exception_data_t exc_data,
+    mach_msg_type_number_t exc_data_count, int *flavor,
+    thread_state_t old_state, mach_msg_type_number_t old_stateCnt,
+    thread_state_t new_state, mach_msg_type_number_t *new_stateCnt) {
+  mach_port_deallocate(mach_task_self(), task_port);
+  mach_port_deallocate(mach_task_self(), thread_port);
+  return KERN_FAILURE;
+}
+
+kern_return_t
+catch_mach_exception_raise(mach_port_t exc_port, mach_port_t thread_port,
+                           mach_port_t task_port, exception_type_t exc_type,
+                           mach_exception_data_t exc_data,
+                           mach_msg_type_number_t exc_data_count) {
+  return KERN_FAILURE;
+}
+
+boolean_t mach_exc_server(mach_msg_header_t *InHeadP,
+                          mach_msg_header_t *OutHeadP);
+}
+
 namespace ds2 {
 namespace Host {
 namespace Darwin {
--- a/Sources/Host/Darwin/Mach.cpp
+++ b/Sources/Host/Darwin/Mach.cpp
@@ -93,7 +93,13 @@ namespace ds2 {
 namespace Host {
 namespace Darwin {
 
+Mach::Mach(ProcessId pid) { _task = getMachTask(pid); }
+
 task_t Mach::getMachTask(ProcessId pid) {
+  if (pid == -1) {
+    return _task;
+  }
+
   task_t self = mach_task_self();
   task_t task;
 
@@ -110,8 +116,9 @@ thread_t Mach::getMachThread(ProcessThreadId const &ptid) {
   mach_msg_type_number_t thread_count;
 
   mach_port_t task = getMachTask(ptid.pid);
-  if (task == TASK_NULL)
+  if (task == TASK_NULL) {
     return THREAD_NULL;
+  }
 
   kern_return_t kret = task_threads(task, &thread_list, &thread_count);
   if (kret != KERN_SUCCESS) {
@@ -132,8 +139,9 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
   mach_vm_size_t curr_bytes_read = 0;
 
   mach_port_t task = getMachTask(ptid.pid);
-  if (task == TASK_NULL)
+  if (task == TASK_NULL) {
     return kErrorProcessNotFound;
+  }
 
   kern_return_t kret =
       mach_vm_read_overwrite((vm_map_t)task, address, length,
@@ -155,7 +163,7 @@ ErrorCode Mach::writeMemory(ProcessThreadId const &ptid, Address const &address,
   // The mach debugging APIs do not allow writing to pages mapped without write
   // permissions; remap them before writing.
   // TODO: Write in multipage ?
-  task_t task = getMachTask(ptid.pid);
+  task_t task = getMachTask();
   if (task == TASK_NULL) {
     return kErrorProcessNotFound;
   }
@@ -221,7 +229,7 @@ ErrorCode Mach::resume(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
                        int signal, Address const &address) {
   kern_return_t kret;
   thread_t thread = getMachThread(ptid);
-  task_t task = getMachTask(ptid.pid);
+  task_t task = getMachTask();
   struct thread_basic_info info;
   unsigned int info_count = THREAD_BASIC_INFO_COUNT;
 
@@ -273,8 +281,8 @@ ErrorCode Mach::resume(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
   return kSuccess;
 }
 
-ErrorCode Mach::getProcessDylbInfo(ProcessId pid, Address &address) {
-  task_t task = getMachTask(pid);
+ErrorCode Mach::getProcessDylbInfo(Address &address) {
+  task_t task = getMachTask();
   if (task == TASK_NULL) {
     return kErrorProcessNotFound;
   }
@@ -292,9 +300,9 @@ ErrorCode Mach::getProcessDylbInfo(ProcessId pid, Address &address) {
   return kSuccess;
 }
 
-ErrorCode Mach::getProcessMemoryRegion(ProcessId pid, Address const &address,
+ErrorCode Mach::getProcessMemoryRegion(Address const &address,
                                        MemoryRegionInfo &region) {
-  task_t task = getMachTask(pid);
+  task_t task = getMachTask();
   if (task == TASK_NULL) {
     return kErrorProcessNotFound;
   }
@@ -362,10 +370,10 @@ Mach::getThreadIdentifierInfo(ProcessThreadId const &ptid,
   return kSuccess;
 }
 
-ErrorCode Mach::setupExceptionChannel(ProcessId pid) {
+ErrorCode Mach::setupExceptionChannel() {
   task_t self = mach_task_self();
 
-  task_t task = getMachTask(pid);
+  task_t task = getMachTask();
   if (task == TASK_NULL) {
     return kErrorProcessNotFound;
   }
--- a/Sources/Host/Darwin/Mach.cpp
+++ b/Sources/Host/Darwin/Mach.cpp
@@ -219,7 +219,58 @@ ErrorCode Mach::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
 
 ErrorCode Mach::resume(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
                        int signal, Address const &address) {
-  return kErrorUnsupported;
+  kern_return_t kret;
+  thread_t thread = getMachThread(ptid);
+  task_t task = getMachTask(ptid.pid);
+  struct thread_basic_info info;
+  unsigned int info_count = THREAD_BASIC_INFO_COUNT;
+
+  if (thread == THREAD_NULL) {
+    return kErrorProcessNotFound;
+  }
+
+  kret = mach_msg(&_reply.hdr, MACH_SEND_MSG | MACH_SEND_TIMEOUT,
+                  _reply.hdr.msgh_size, 0, MACH_PORT_NULL, 100, MACH_PORT_NULL);
+  if (kret != KERN_SUCCESS) {
+    DS2LOG(Error, "Failed to reply");
+    return kErrorUnknown;
+  }
+
+  kret = thread_info((thread_t)thread, THREAD_BASIC_INFO, (thread_info_t)&info,
+                     &info_count);
+  if (kret != KERN_SUCCESS) {
+    return kErrorProcessNotFound;
+  }
+
+  for (int i = 0; i < info.suspend_count; i++) {
+    DS2LOG(Debug, "[1] Resume pid:%d", ptid.pid);
+    if (thread_resume(thread) != KERN_SUCCESS)
+      continue;
+  }
+
+  kret = thread_info((thread_t)thread, THREAD_BASIC_INFO, (thread_info_t)&info,
+                     &info_count);
+  if (kret != KERN_SUCCESS) {
+    return kErrorProcessNotFound;
+  }
+  DS2ASSERT(info.suspend_count == 0);
+
+  struct task_basic_info tinfo;
+  mach_msg_type_number_t count = TASK_BASIC_INFO_COUNT;
+  kret = ::task_info(task, TASK_BASIC_INFO, (task_info_t)&tinfo, &count);
+  if (kret != KERN_SUCCESS) {
+    return Platform::TranslateKernError(kret);
+  }
+
+  if (tinfo.suspend_count != 0) {
+    DS2LOG(Debug, "Resume task");
+    kret = task_resume(task);
+    if (kret != KERN_SUCCESS) {
+      return kErrorProcessNotFound;
+    }
+  }
+
+  return kSuccess;
 }
 
 ErrorCode Mach::getProcessDylbInfo(ProcessId pid, Address &address) {
--- a/Sources/Host/Darwin/Mach.cpp
+++ b/Sources/Host/Darwin/Mach.cpp
@@ -28,6 +28,16 @@
 // to these system can't be register, it's hardlink. So we need to compile
 // these function to have access to the exception system. Also it's C...
 extern "C" {
+
+static struct {
+  task_t task;
+  bool isFilled;
+  thread_t thread;
+  int excType;
+  int data_count;
+  mach_exception_data_type_t data[1024];
+} g_exec_state;
+
 kern_return_t catch_mach_exception_raise_state(
     mach_port_t exc_port, exception_type_t exc_type,
     const mach_exception_data_t exc_data, mach_msg_type_number_t exc_data_count,
@@ -53,6 +63,25 @@ catch_mach_exception_raise(mach_port_t exc_port, mach_port_t thread_port,
                            mach_port_t task_port, exception_type_t exc_type,
                            mach_exception_data_t exc_data,
                            mach_msg_type_number_t exc_data_count) {
+
+  if (1) { // g_exec_state.task == task_port) {
+    kern_return_t kret;
+
+    kret = thread_suspend(thread_port);
+    if (kret != KERN_SUCCESS) {
+      DS2LOG(Error, "Unable to suspend the thread: %s",
+             mach_error_string(kret));
+      return KERN_FAILURE;
+    }
+
+    g_exec_state.thread = thread_port;
+    g_exec_state.excType = exc_type;
+    g_exec_state.isFilled = 1;
+    g_exec_state.data_count = exc_data_count;
+    ::memcpy(g_exec_state.data, exc_data,
+             exc_data_count * sizeof(mach_exception_data_type_t));
+    return KERN_SUCCESS;
+  }
   return KERN_FAILURE;
 }
 
@@ -280,6 +309,74 @@ Mach::getThreadIdentifierInfo(ProcessThreadId const &ptid,
   }
 
   return kSuccess;
+}
+
+ErrorCode Mach::setupExceptionChannel(ProcessId pid) {
+  task_t self = mach_task_self();
+
+  task_t task = getMachTask(pid);
+  if (task == TASK_NULL) {
+    return kErrorProcessNotFound;
+  }
+
+  kern_return_t kret =
+      mach_port_allocate(self, MACH_PORT_RIGHT_RECEIVE, &_exc_port);
+  if (kret != KERN_SUCCESS) {
+    return Platform::TranslateKernError(kret);
+  }
+
+  kret = mach_port_insert_right(self, _exc_port, _exc_port,
+                                MACH_MSG_TYPE_MAKE_SEND);
+  if (kret != KERN_SUCCESS) {
+    return Platform::TranslateKernError(kret);
+  }
+
+  kret = task_set_exception_ports(
+      task, EXC_MASK_BREAKPOINT | EXC_MASK_SOFTWARE, _exc_port,
+      EXCEPTION_DEFAULT | MACH_EXCEPTION_CODES, THREAD_STATE_NONE);
+
+  return Platform::TranslateKernError(kret);
+}
+
+ErrorCode Mach::readException(MachExcStatus &status, bool timeout,
+                              thread_t *thread) {
+  union {
+    mach_msg_header_t hdr;
+    char payload[1024];
+  } msg;
+
+  mach_msg_option_t opts = MACH_RCV_MSG | MACH_RCV_INTERRUPT;
+  if (timeout) {
+    opts |= MACH_RCV_TIMEOUT;
+  }
+
+  kern_return_t kret = mach_msg(&msg.hdr, opts, 0, sizeof(msg.payload),
+                                _exc_port, 10000, MACH_PORT_NULL);
+  if (kret != MACH_MSG_SUCCESS) {
+    return Platform::TranslateKernError(kret);
+  }
+
+  g_exec_state.isFilled = 0;
+
+  if (!::mach_exc_server(&msg.hdr, &_reply.hdr)) {
+    return readException(status, timeout, thread);
+  } else if (g_exec_state.isFilled == 0) {
+    return readException(status, timeout, thread);
+  }
+
+  if (thread) {
+    *thread = g_exec_state.thread;
+  }
+
+  status.type = g_exec_state.excType;
+  status.subtype = g_exec_state.data[0];
+  status.data = g_exec_state.data[1];
+
+  return kSuccess;
+}
+
+bool Mach::exceptionIsFromThread(ProcessThreadId const &ptid) {
+  return g_exec_state.thread == getMachThread(ptid);
 }
 }
 }
--- a/Sources/Host/Darwin/PTrace.cpp
+++ b/Sources/Host/Darwin/PTrace.cpp
@@ -31,9 +31,23 @@ PTrace::PTrace() : _privateData(nullptr) {}
 
 PTrace::~PTrace() { doneCPUState(); }
 
+ErrorCode PTrace::traceMe(bool disableASLR) {
+  ErrorCode error = super::traceMe(disableASLR);
+  if (error != kSuccess) {
+    return error;
+  }
+
+  if (wrapPtrace(PT_SIGEXC, 0, nullptr, nullptr) < 0) {
+    return Platform::TranslateError();
+  }
+
+  return kSuccess;
+}
+
 ErrorCode PTrace::traceThat(ProcessId pid) {
-  if (pid <= 0)
+  if (pid <= 0) {
     return kErrorInvalidArgument;
+  }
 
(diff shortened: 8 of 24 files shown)


kokarez:
With this last version, breakpoints, continue, read, step work.

Headers/DebugServer2/Host/Darwin/Mach.h
```diff
@@ -0,0 +1,67 @@
...
+
+namespace ds2 {
+namespace Host {
+namespace Darwin {
+
+struct MachPrivateData;
+
+class Mach {
+public:
+  Mach();
```
sas: Get rid of the empty constructor.

Headers/DebugServer2/Host/Darwin/Mach.h
```diff
@@ -0,0 +1,67 @@
...
+  virtual ~Mach();
```
sas: `virtual ~Mach() = default;`

Sources/Host/Darwin/Mach.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <sys/types.h>
+#include <mach/mach.h>
+#include <mach/mach_vm.h>
+#include <mach/thread_info.h>
+
+namespace ds2 {
+namespace Host {
+namespace Darwin {
+
+Mach::Mach() {}
```
sas: And you can get rid of these two empty ctor and dtor.

Headers/DebugServer2/Host/Darwin/Mach.h
```diff
@@ -0,0 +1,92 @@
...
+#include <sys/types.h>
+#include <mach/mach.h>
+#include <mach/mach_vm.h>
+#include <mach/thread_info.h>
+
+namespace ds2 {
+namespace Host {
+namespace Darwin {
+
+struct MachPrivateData;
```
sas: I think this is useless.

Sources/Host/Darwin/LibProc.cpp
```diff
@@ -55,7 +58,23 @@ void LibProc::EnumerateProcesses(
...
+  err = mach.getThreadIdentifierInfo(pid, tid, &threadId);
+  if (err != kSuccess)
+    return dft;
+
+  res = proc_pidinfo(pid, PROC_PIDTHREADINFO, threadId.thread_handle, &ti,
+                     sizeof(ti));
+  if (res <= 0)
+    return dft;
+
+  return std::string(ti.pth_name);
```
sas: The constructor for `std::string` is implicit, I think you can just do `return ti.pth_name;`

Sources/Host/Darwin/LibProc.cpp
```diff
@@ -55,7 +58,23 @@ void LibProc::EnumerateProcesses(
 }
 
 std::string LibProc::GetThreadName(ProcessId pid, ThreadId tid) {
-  return "<unknown>";
+  std::string dft("<unknown-thread>");
```
sas: Can you use `<unknown>` instead of `<unknown-thread>`? I just changed our default strings to be more uniform.

Sources/Host/Darwin/LibProc.cpp
```diff
@@ -55,7 +58,23 @@ void LibProc::EnumerateProcesses(
...
+  thread_identifier_info_data_t threadId;
+  struct proc_threadinfo ti;
+  ErrorCode err;
+  Mach mach(pid);
```
sas: Sucks a little bit that we have to create a whole new `Mach` object here just to get a thread name.

Headers/DebugServer2/Target/Darwin/MachOProcess.h
```diff
@@ -21,6 +22,7 @@ class MachOProcess : public ds2::Target::POSIX::Process {
 protected:
   std::string _auxiliaryVector;
   Address _sharedLibraryInfoAddress;
+  Host::Darwin::Mach _mach;
```
sas: I think this (and the getter that goes with it on line 40) should be in DarwinProcess instead. The Mach interface is not associated with the binary format, but rather the OS you're running on.

Also, I might get rid of `ELFProcess` and `MachOProcess` at some point soon, as they both seem like questionable design, and we can integrate their functionality in a mixin I think.

Sources/Host/Darwin/PTrace.cpp
```diff
@@ -57,16 +58,8 @@ ErrorCode PTrace::kill(ProcessThreadId const &ptid, int signal) {
...
-  ErrorCode err = readMemory(ptid, address, buf, length, count);
-  if (err != kSuccess)
-    return err;
-
-  if (strnlen(buf, length) == length)
-    return kErrorNameTooLong;
-
-  str = std::string(buf);
-  return kSuccess;
+  DS2BUG("not implemented");
```
sas: I'd think that calling `PTrace::readString` on a Darwin build is a nasty bug, so something more specific than `"not implemented"` would be good. E.g.: `DS2BUG("impossible to use ptrace to %s on Darwin", __FUNCTION__);`

Sources/Host/Darwin/PTrace.cpp
```diff
@@ -57,16 +58,8 @@ ErrorCode PTrace::kill(ProcessThreadId const &ptid, int signal) {
...
+  return kErrorUnsupported;
```
sas: `DS2BUG` is marked `__attribute__((__noreturn__))` so you don't need the `return kErrorUnsupported;` here.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -0,0 +1,157 @@
...
+    return THREAD_NULL;
+
+  kret = task_threads(task, &thread_list, &thread_count);
+  if (kret != KERN_SUCCESS) {
+    DS2LOG(Error, "Unable to retrieve the threads: %s",
+           mach_error_string(kret));
+    return THREAD_NULL;
+  }
+
+  /* need to have a real way to get the correct thread */
```
sas: Can you use `//` style comments instead of `/* */`?

Sources/Host/Darwin/X86_64/MachX86_64.cpp
```diff
@@ -0,0 +1,118 @@
...
+  kret = thread_get_state(thread, x86_THREAD_STATE, (thread_state_t)&state,
+                          &stateCount);
+  if (kret != KERN_SUCCESS) {
+    DS2LOG(Error, "Fail to get the reg of the pid: %s",
+           mach_error_string(kret));
+    return kErrorInvalidArgument;
+  }
+
+  ds2_state.is32 = false;
+  ds2_state.state64.gp.rax = state.uts.ts64.__rax;
```
sas: For this big block of registers, look at `user_to_state64` and `state64_to_user`. If the macros are too much of a pain to configure/use, I'm fine with checking in the current version of the register copy; I'll make the change myself later on.

Sources/Host/Darwin/X86_64/PTraceX86_64.cpp
```diff
@@ -44,34 +44,13 @@ void PTrace::doneCPUState() { delete _privateData; }
...
-
-  ErrorCode error = ptidToPid(ptid, pid);
-  if (error != kSuccess)
-    return error;
-
-  initCPUState(pid);
-  DS2BUG("not implemented");
-
-  return kSuccess;
+  return kErrorUnsupported;
```
sas: These should use `DS2BUG` too.

Sources/Host/FreeBSD/X86_64/PTraceX86_64.cpp
```diff
@@ -9,8 +9,8 @@
 // PATENTS file in the same directory.
```
sas: You shouldn't touch any Linux or FreeBSD PTrace file in this commit. I think you're using an incorrect version of clang-format. You want version 3.7 or 3.8 (whichever one we're using on Travis) IIRC.

Sources/Target/Darwin/Process.cpp
```diff
@@ -311,6 +314,42 @@ ErrorCode Process::getMemoryRegionInfo(Address const &address,
...
+    return kErrorNameTooLong;
+
+  str = std::string(buf);
+  return kSuccess;
+}
+
+ErrorCode Process::readMemory(Address const &address, void *data, size_t length,
+                              size_t *count) {
+  if (_currentThread == nullptr)
+    return super::readMemory(address, data, length, count);
```
sas: This is going to end up calling `PTrace::readMemory`. Not sure you want that.

Sources/Target/Darwin/Process.cpp
```diff
@@ -311,6 +314,42 @@ ErrorCode Process::getMemoryRegionInfo(Address const &address,
...
+
+  return mach().readMemory(_currentThread->tid(), address, data, length, count);
+}
+
+ErrorCode Process::writeMemory(Address const &address, void const *data,
+                               size_t length, size_t *count) {
+  if (_currentThread == nullptr)
+    return super::writeMemory(address, data, length, count);
```
sas: Same here.

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,8 +32,39 @@ namespace Darwin {
...
 using Host::Darwin::LibProc;
 
-Thread::Thread(Process *process, ThreadId tid)
-    : super(process, tid), _lastSyscallNumber(-1) {}
+Thread::Thread(ds2::Target::Process *process, ThreadId tid)
+    : super(process, tid), _lastSyscallNumber(-1) {
+  //
+  // Initially the thread is stopped.
+  //
+  _state = kStopped;
```
sas: The body of this constructor is useless. _state is already set to kStopped by `POSIX::Thread::Thread` and `_lastSyscallNumber` is already set to `-1` by the initializer list.

Sources/Target/Linux/ARM/ProcessARM.cpp
```diff
@@ -8,9 +8,9 @@
 // PATENTS file in the same directory.
```
sas: Same here, you shouldn't touch these headers.

Support/Testing/Travis/script.sh
```diff
@@ -71,6 +71,7 @@ check_dirty() {
     for f in "${dirty[@]}"; do
       echo "dirty: $f"
     done
+    git diff | cat
```
sas: no 😆

Support/Testing/Travis/keychain.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
```
sas: I think this script is fine for now, but putting keys in the repo is not the proper way of doing this. We can fix it later.

Sources/Support/POSIX/Stringify.cpp
```diff
@@ -262,6 +262,14 @@ char const *Stringify::Errno(int error) {
 
 char const *Stringify::Ptrace(int code) {
   switch (code) {
+#if defined(OS_DARWIN)
```
sas: Put these at the end, and do `#elif defined(OS_DARWIN) || defined(OS_FREEBSD)`.

These values work for FreeBSD too so we might as well define them here.

sas:
I reviewed the first big patch (Mach system). Overall the logic is good but there are a few things that need to be fixed before it goes in.

I also reviewed the last two tiny patches, the travis stuff, and the Stringify patch.
When you fix the Stringify patch I think we can submit it right away. The travis file needs to be clean up a little bit (but you know about it already 😄 ). I think it's worth taking care of the travis one ASAP so that each subsequent patch/CR is build on Darwin as well.

Sources/Host/Darwin/LibProc.cpp
kokarez: Yea. I did that du to bad dependency. I mean I don't since it's a huge problem, a mach object is here to describe a process, so you will need to create a new one each time you want to get specific info.

Headers/DebugServer2/Target/Darwin/MachOProcess.h
kokarez: It's here because it's at the lower level in the hierarchy, since we need it in MachOProcess.cpp. When you delete these file, you can move it up.

Sources/Host/Darwin/PTrace.cpp
kokarez: 👍

Sources/Host/FreeBSD/X86_64/PTraceX86_64.cpp
kokarez: Yea that's unfortunate. I will delete that.

Sources/Support/POSIX/Stringify.cpp
kokarez: I just check the man PTrace on freebsd and it should work. Unfortunately freebsd build doesn't compile at this moment and I can't test :D So I will do a separate commit for freebsd PTrace stringify.

.travis.yml
```diff
@@ -9,42 +9,80 @@
...
     - master
-matrix:
-  fast_finish: true
 
-env:
-  - TARGET="Style"
-  - TARGET="Registers"
+matrix:
+  include:
+    - os: osx
```
sas: Testing all these versions of Xcode is not super useful. Having only 7 and 7.3 is good enough for now probably.

sas: Also, move the Darwin builds at the end of the block, after the LLDB unit test ones please.

.travis.yml
```diff
@@ -9,42 +9,80 @@
...
+      osx_image: xcode7.3
+      env: TARGET="Darwin 7.3"
```
sas: Use `TARGET=Darwin` for all of these, the xcode version shouldn't appear in the `TARGET` env var.

.travis.yml
```diff
@@ -9,42 +9,80 @@
...
+      env: TARGET="Linux-X86_64"   LLDB_TESTS="all"  CLANG="0" COVERAGE="1"
+    - os: linux
+      env: TARGET="Linux-X86_64"   LLDB_TESTS="all"  CLANG="0" RELEASE="1"
+    - os: linux
+      env: TARGET="Linux-X86_64"   LLDB_TESTS="all"  CLANG="1" COVERAGE="1"
+    - os: linux
+      env: TARGET="Linux-X86_64"   LLDB_TESTS="all"  CLANG="1" RELEASE="1"
+
+before_install:
+  - ./Support/Testing/Travis/keychain.sh
```
sas: We don't need signature-related stuff for now.

Support/Testing/Travis/after_success.sh
```diff
@@ -11,6 +11,10 @@
 
 set -eu
 
+if [[ "${TRAVIS_OS_NAME-}" = "osx" ]]; then
```
sas: Can you use the `TARGET` variable instead of `TRAVIS_OS_NAME`? We try to make all the scripts in `Support/Testing/Travis` usable from our actual dev hosts (although for `after_success.sh` it really doesn't matter that much).

Support/Testing/Travis/before-install.py
```diff
@@ -10,12 +10,18 @@
...
 
 import os
+import sys
 from subprocess import check_call
 
 repositories = []
 keys = []
 
 target = os.getenv('TARGET')
+travis_os = os.getenv('TRAVIS_OS_NAME')
```
sas: Same here, you don't need `travis_os`, you can just use `target` and do `if target == 'Darwin':`.

Support/Testing/Travis/install.py
```diff
@@ -75,7 +79,10 @@
     packages.append('realpath')
 
 if len(packages) > 0:
-    if "CentOS Linux" in platform.linux_distribution():
+    if travis_os == 'osx':
+        check_call('brew install "%s" || true' % '" "'.join(packages), shell=True)
```
sas: I know why you're doing this `|| true` thing with brew but I wish there was a better way to take care of this :/
Can you add a tiny comment to explain why so that next person to read this code gets it?

Support/Testing/Travis/keychain.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
sas: This along with the `ds2_codesign.p12` file should be removed from this commit.

sas:
Committed the Mach system patch as 1c296bb1f9f577490937b8afff831728eea8dfd8.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -109,6 +109,28 @@ ErrorCode Mach::resume(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
   return kErrorUnsupported;
 }
 
+ErrorCode Mach::getProcessDylbInfo(pid_t pid, task_dyld_info_t dyldInfo) {
+  task_t task;
+
+  task = getMachTask(pid);
+  if (task == TASK_NULL) {
+    DS2LOG(Debug, "Unable to get the task");
```
sas: `s/Unable/unable/`

Sources/Host/Darwin/Mach.cpp
```diff
@@ -109,6 +109,28 @@ ErrorCode Mach::resume(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
...
+    return kErrorProcessNotFound;
+  }
+
+  kern_return_t kret;
+  mach_msg_type_number_t count = TASK_DYLD_INFO_COUNT;
+
+  kret = task_info(task, TASK_DYLD_INFO, (task_info_t)dyldInfo, &count);
+  if (kret) {
+    DS2LOG(Debug, "Fail to retreive the dyld info: %s",
```
sas: s/Fail/failed/

Sources/Target/Darwin/MachOProcess.cpp
```diff
@@ -14,10 +14,17 @@
 #include <dirent.h>
 #include <limits>
```
sas: No blank line, sort all includes together.

Sources/Target/Darwin/MachOProcess.cpp
```diff
@@ -109,8 +114,16 @@ ErrorCode MachOProcess::updateAuxiliaryVector() { return kSuccess; }
 // Retrieves the shared library info address pointer.
 //
 ErrorCode MachOProcess::getSharedLibraryInfoAddress(Address &address) {
-  DS2BUG("not implemented");
-  return kErrorUnsupported;
+  task_dyld_info_data_t dyldInfo;
+  ErrorCode err;
+
+  err = mach().getProcessDylbInfo(_info.pid, &dyldInfo);
```
sas:
```
ErrorCode error = mach().get...;
if (error != kSuccess) {
    return error;
}
```

Sources/Target/Darwin/MachOProcess.cpp
```diff
@@ -14,10 +14,17 @@
 #include <dirent.h>
 #include <limits>
 
+#include <mach/mach.h>
```
sas: No blank line, sort all includes together.

Sources/Host/Darwin/Platform.cpp
```diff
@@ -10,21 +10,74 @@
...
 namespace Host {
 namespace Darwin {
 
 char const *Platform::GetOSTypeName() { return "macosx"; }
 
-char const *Platform::GetOSVendorName() { return "unknown"; }
+char const *Platform::GetOSVersion() {
+  static char version[9] = {0};
+
+  /*
```
sas: Use `//` style comments.

Sources/Host/Darwin/Platform.cpp
```diff
@@ -10,21 +10,74 @@
...
+   *
+   * TODO: Found a better way to do that
+   */
+
+  if (version[0] == '\0') {
+    std::string line;
+    std::ifstream versionFile(
+        "/System/Library/CoreServices/SystemVersion.plist");
+
+    if (versionFile.is_open()) {
```
sas: Prefer early returns:
```
if (!versionFile.is_open()) {
    return version;
}

[...]
```

Sources/Host/Darwin/Platform.cpp
```diff
@@ -10,21 +10,74 @@
...
 
 char const *Platform::GetOSTypeName() { return "macosx"; }
 
-char const *Platform::GetOSVendorName() { return "unknown"; }
+char const *Platform::GetOSVersion() {
+  static char version[9] = {0};
+
+  /*
+   * Mac doesn't offer a C/C++ call to get the OS version. SystemVersion.plist
+   * is use buy officiel application to get the version, so I think it's ok
```
sas: s/is use buy officiel application/is used by official applications/

s/so I think it's ok to read it//

s/An xml parser will be nice, but it will also add a lot of deps/An XML parser would be nice but would add a heavy dependency/

Sources/Host/Darwin/Platform.cpp
```diff
@@ -10,21 +10,74 @@
...
+   * to read it. An xml parser will be nice, but it will also add a lot of
+   * deps to read one static value.
+   *
+   * TODO: Found a better way to do that
```
sas: I think you're right, and there probably isn't a better way to do that, so you can remove this TODO.

Sources/Host/Darwin/Platform.cpp
```diff
@@ -10,21 +10,74 @@
...
+   */
+
+  if (version[0] == '\0') {
+    std::string line;
+    std::ifstream versionFile(
+        "/System/Library/CoreServices/SystemVersion.plist");
+
+    if (versionFile.is_open()) {
+      while (getline(versionFile, line)) {
```
sas: Use `std::getline`. Makes it more obvious that this isn't the C function call.

Sources/Host/Darwin/Platform.cpp
```diff
@@ -10,21 +10,74 @@
...
+        size_t start = 0;
+        size_t end = 0;
+        size_t len = 0;
+
+        if (line.find("ProductVersion") == std::string::npos)
+          continue;
+
+        if (!getline(versionFile, line))
```
sas: Ditto.

Sources/Host/Darwin/Platform.cpp
```diff
@@ -10,21 +10,74 @@
...
+        end = line.find("</string>");
+
+        if (start == std::string::npos || end == std::string::npos)
+          break;
+
+        start += 8;
+
+        len = (end - start) > 8 ? 8 : end - start;
+
+        memcpy(version, line.substr(start, end).c_str(), len);
```
sas: std::memcpy

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+  if (count)
+    *count = curr_bytes_read;
 
   return kSuccess;
 }
 
 ErrorCode Mach::writeMemory(ProcessThreadId const &ptid, Address const &address,
                             void const *buffer, size_t length, size_t *count) {
-  return kErrorUnsupported;
+  /*
```
sas: `//` style comments.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+   * We want to write in the process memory, but most of the write will be
```
sas: The mach debugging APIs do not allow writing to pages mapped without write permissions; remap them before writing.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+   * 1) get current permissions
+   * 2) set write permissions
+   * 3) write
+   * 4) restore permissions
+   *
+   * TODO: Write in multipage ?
+   */
+  task_t task;
+
+  task = getMachTask(ptid.pid);
```
sas: `task_t task = getMachTask(ptid.pid);`

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+
+  mach_msg_type_number_t infoCount = VM_REGION_SUBMAP_INFO_COUNT_64;
+  vm_region_submap_info_data_64_t regionInfo;
+  mach_vm_address_t start = address.value();
+  mach_vm_size_t size;
+  kern_return_t kret;
+  ErrorCode err = kSuccess;
+  natural_t depth = 1024;
+
+  kret =
```
sas: Ditto.

Sources/Host/Darwin/X86_64/MachX86_64.cpp
```diff
@@ -94,7 +94,7 @@ ErrorCode Mach::readCPUState(ProcessThreadId const &ptid,
 
 ErrorCode Mach::writeCPUState(ProcessThreadId const &ptid,
                               ProcessInfo const &pinfo,
-                              Architecture::CPUState const &state) {
+                              Architecture::CPUState const &ds2_state) {
```
sas: You should keep this named `state` and name your local variable `threadState`, or `context` or something like that.

Sources/Host/Darwin/X86_64/MachX86_64.cpp
```diff
@@ -106,10 +106,46 @@ ErrorCode Mach::writeCPUState(ProcessThreadId const &ptid,
...
+  state.uts.ts64.__r13 = ds2_state.state64.gp.r13;
+  state.uts.ts64.__r14 = ds2_state.state64.gp.r14;
+  state.uts.ts64.__r15 = ds2_state.state64.gp.r15;
+  state.uts.ts64.__rip = ds2_state.state64.gp.rip;
+  state.uts.ts64.__cs = ds2_state.state64.gp.cs & 0xffff;
+  state.uts.ts64.__fs = ds2_state.state64.gp.fs & 0xffff;
+  state.uts.ts64.__gs = ds2_state.state64.gp.gs & 0xffff;
+  state.uts.ts64.__rflags = ds2_state.state64.gp.eflags;
+
+  kret = thread_set_state(thread, x86_THREAD_STATE, (thread_state_t)&state,
```
sas: `kern_return_t kret = thread_set_state(...`

Sources/Host/Darwin/X86_64/MachX86_64.cpp
```diff
@@ -106,10 +106,46 @@ ErrorCode Mach::writeCPUState(ProcessThreadId const &ptid,
...
+                          stateCount);
+  if (kret != KERN_SUCCESS) {
+    DS2LOG(Error, "Fail to set the reg of the pid: %s",
```
sas: no need.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -146,7 +146,15 @@ ErrorCode Mach::writeMemory(ProcessThreadId const &ptid, Address const &address,
 }
 
 ErrorCode Mach::suspend(ProcessThreadId const &ptid) {
-  return kErrorUnsupported;
+  thread_t thread;
+
+  thread = getMachThread(ptid.pid, ptid.tid);
```
sas: `thread_t thread = getMachThread(...`

Sources/Host/Darwin/Mach.cpp
```diff
@@ -146,7 +146,15 @@ ErrorCode Mach::writeMemory(ProcessThreadId const &ptid, Address const &address,
...
+  if (thread == THREAD_NULL)
+    return kErrorProcessNotFound;
```
sas: Braces.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -146,7 +146,15 @@ ErrorCode Mach::writeMemory(ProcessThreadId const &ptid, Address const &address,
...
+
+  thread_suspend(thread);
```
sas: Check return value.

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,7 +33,88 @@ namespace Darwin {
 using Host::Darwin::LibProc;
 
 Thread::Thread(ds2::Target::Process *process, ThreadId tid)
-    : super(process, tid), _lastSyscallNumber(-1) {}
+    : super(process, tid), _lastSyscallNumber(-1) {
+  //
+  // Initially the thread is stopped.
+  //
+  _resumeWithPtrace = true;
```
sas: Use initializer list for this and add a comment that explains why the first resume should be done with ptrace.

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,7 +33,88 @@ namespace Darwin {
...
+}
+
+ErrorCode Thread::suspend() {
+  ErrorCode error = kSuccess;
+  if (_state == kRunning) {
+    error = process()->mach().suspend(ProcessThreadId(process()->pid(), tid()));
+    if (error != kSuccess)
+      return error;
+    updateStopInfo(1407);
```
sas: What is `1407`?

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,7 +33,88 @@ namespace Darwin {
...
 Thread::Thread(ds2::Target::Process *process, ThreadId tid)
-    : super(process, tid), _lastSyscallNumber(-1) {}
+    : super(process, tid), _lastSyscallNumber(-1) {
+  //
+  // Initially the thread is stopped.
+  //
+  _resumeWithPtrace = true;
+}
+
+ErrorCode Thread::suspend() {
```
sas: Shouldn't you set _resumeWithPtrace to either true or false in this function?

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,7 +33,88 @@ namespace Darwin {
...
+  }
+  return error;
+}
+
+ErrorCode Thread::resume(int signal, Address const &address) {
+  ErrorCode error = kSuccess;
+  if (_state == kStopped || _state == kStepped) {
+    if (signal == 0) {
+      switch (_stopInfo.signal) {
+      case SIGCHLD:
```
sas: You can remove the special handling of these three signals. We used to have that in the Linux implementation but there's no use for it.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+  natural_t depth = 1024;
+
+  kret =
+      mach_vm_region_recurse(task, &start, &size, &depth,
+                             (vm_region_recurse_info_t)&regionInfo, &infoCount);
+  if (kret != KERN_SUCCESS)
+    return kErrorInvalidAddress;
+
+  /*
+   * The combinason of VM_PROT_COPY/VM_PROT_READ is need to override the
```
sas: no need.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+  mach_vm_size_t size;
+  kern_return_t kret;
+  ErrorCode err = kSuccess;
+  natural_t depth = 1024;
+
+  kret =
+      mach_vm_region_recurse(task, &start, &size, &depth,
+                             (vm_region_recurse_info_t)&regionInfo, &infoCount);
+  if (kret != KERN_SUCCESS)
+    return kErrorInvalidAddress;
```
sas: braces (also for the rest of the file)

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+  if (task == TASK_NULL) {
+    return kErrorProcessNotFound;
+  }
+
+  mach_msg_type_number_t infoCount = VM_REGION_SUBMAP_INFO_COUNT_64;
+  vm_region_submap_info_data_64_t regionInfo;
+  mach_vm_address_t start = address.value();
+  mach_vm_size_t size;
+  kern_return_t kret;
+  ErrorCode err = kSuccess;
```
sas: `ErrorCode error = kSuccess;`

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+  kret = mach_vm_protect(task, address.value(), length, FALSE,
+                         VM_PROT_WRITE | VM_PROT_COPY | VM_PROT_READ);
+  if (kret != KERN_SUCCESS)
+    return kErrorInvalidAddress;
+
+  kret = mach_vm_write((vm_map_t)task, address.value(), (vm_offset_t)buffer,
+                       length);
+  if (kret != KERN_SUCCESS)
+    err = kErrorUnknown;
+  else if (count)
```
sas: `count != nullptr`

Sources/Host/Darwin/Mach.cpp
```diff
@@ -80,14 +80,69 @@ ErrorCode Mach::readMemory(ProcessThreadId const &ptid, Address const &address,
...
+    *count = length;
+
+  kret = mach_vm_protect(task, address.value(), length, FALSE,
+                         regionInfo.protection);
+  if (kret != KERN_SUCCESS)
```
sas: I think here you should see if `err` is already set to an error code (i.e.: if the previous call to `mach_vm_write` failed. Something like `if (kret != KERN_SUCCESS && err == kSuccess)` should be better, no?

Sources/Host/Darwin/Mach.cpp
```diff
@@ -195,6 +195,26 @@ ErrorCode Mach::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
   return kErrorUnsupported;
 }
 
+ErrorCode Mach::enableHardwareSingleStep(ProcessThreadId const &ptid,
```
sas: This shouldn't be here.

Sources/Host/Darwin/PTrace.cpp
```diff
@@ -37,6 +37,8 @@ ErrorCode PTrace::traceMe(bool disableASLR) {
 
   wrapPtrace(PT_SIGEXC, 0, nullptr, nullptr);
 
+  sleep(1);
```
sas: Nope.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -246,7 +246,60 @@ ErrorCode Mach::enableHardwareSingleStep(ProcessThreadId const &ptid,
...
+    return kErrorProcessNotFound;
+  }
+
+  if (tinfo.suspend_count != 0) {
+    DS2LOG(Debug, "Resume task");
+    kret = task_resume(task);
+    if (kret != KERN_SUCCESS) {
+      DS2ASSERT(false);
+      return kErrorProcessNotFound;
+      ;
```
sas: What's this?

Sources/Host/Darwin/Mach.cpp
kokarez: Should not be name like that too :) will rewrite resume/step code.

Sources/Host/Darwin/PTrace.cpp
```diff
@@ -31,6 +31,15 @@ PTrace::PTrace() : _privateData(nullptr) {}
 
 PTrace::~PTrace() { doneCPUState(); }
 
+ErrorCode PTrace::traceMe(bool disableASLR) {
+  if (wrapPtrace(PT_TRACE_ME, 0, nullptr, nullptr) < 0)
```
sas: Instead of doing this, call `super::traceMe`.

Sources/Host/Darwin/PTrace.cpp
```diff
@@ -31,6 +31,15 @@ PTrace::PTrace() : _privateData(nullptr) {}
...
+    return Platform::TranslateError();
+
+  wrapPtrace(PT_SIGEXC, 0, nullptr, nullptr);
```
sas: Check return code.

Sources/Host/Darwin/X86_64/MachX86_64.cpp
```diff
@@ -22,8 +22,6 @@
 #include <sys/uio.h>
 #include <sys/user.h>
 
-#define super ds2::Host::POSIX::Mach
```
sas: Yeah I don't know how this was committed in the first place 😆 

Should I submit this commit right away?

Sources/Host/Darwin/Mach.cpp
```diff
@@ -24,6 +24,39 @@
 #include <mach/thread_info.h>
 #include <sys/types.h>
 
+extern "C" {
```
sas: I don't understand what this block of code is trying to achieve. Can you put a comment explaining it?

CMakeLists.txt
```diff
@@ -65,6 +65,16 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     set(ARCH_NAME "X86")
   endif ()
 endif ()
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
+  set(GENERATED_MACH_SOURCES
+      ${CMAKE_CURRENT_BINARY_DIR}/mach_excServer.c
+      ${CMAKE_CURRENT_BINARY_DIR}/mach_excUser.c
+  )
+  add_custom_command(OUTPUT ${GENERATED_MACH_SOURCES}
+    COMMAND mig ${CMAKE_SOURCE_DIR}/Sources/Host/Darwin/mach_gen.defs
```
sas: nit: You should name these files something like `macGen.defs` instead of mach_gen.defs`.

Sources/Target/Darwin/X86_64/ThreadX86_64.cpp
```diff
@@ -0,0 +1,46 @@
...
+#define __DS2_LOG_CLASS_NAME__ "Target::Thread"
+
+#include "DebugServer2/Target/Darwin/Thread.h"
+#include "DebugServer2/Architecture/CPUState.h"
+#include "DebugServer2/Host/Darwin/LibProc.h"
+#include "DebugServer2/Host/Darwin/PTrace.h"
+#include "DebugServer2/Target/Process.h"
+#include "DebugServer2/Utils/Log.h"
+
+#include <cerrno>
```
sas: I don't think you need any of these system includes.

Sources/Target/Darwin/X86_64/ThreadX86_64.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <sys/wait.h>
+
+#define super ds2::Target::POSIX::Thread
+
+namespace ds2 {
+namespace Target {
+namespace Darwin {
+
+ErrorCode Thread::step(int signal, Address const &address) {
+  // Windows doesn't have a dedicated single-step call. We have to set the
```
sas: `s/Windows/Darwin/`

Sources/Target/Darwin/X86_64/ThreadX86_64.cpp
```diff
@@ -0,0 +1,46 @@
...
+//
+
+#define __DS2_LOG_CLASS_NAME__ "Target::Thread"
+
+#include "DebugServer2/Target/Darwin/Thread.h"
+#include "DebugServer2/Architecture/CPUState.h"
+#include "DebugServer2/Host/Darwin/LibProc.h"
+#include "DebugServer2/Host/Darwin/PTrace.h"
+#include "DebugServer2/Target/Process.h"
+#include "DebugServer2/Utils/Log.h"
```
sas: `LibProc.h`, `PTrace.h`, `Log.h`, and `Process.h` do not seem to be required here.

sas:
Committed 92e145629e938acd3d6d79d38c97debeaaa9b01f.

Sources/Target/Darwin/X86_64/ThreadX86_64.cpp
kokarez: Yea :) I've already clean all these include in the separate PR.

Sources/Target/Darwin/X86_64/ThreadX86_64.cpp
kokarez: Same, This is fix in the separate PR.

Sources/Target/Darwin/Thread.cpp
kokarez: I did move it in process::init and it works! Nice

Sources/Host/Darwin/Mach.cpp
```diff
@@ -24,6 +24,42 @@
 #include <mach/thread_info.h>
 #include <sys/types.h>
 
+// Mach exception are manage by generated code, not a lib. Also the callback
```
sas: s/manage/managed/

Headers/DebugServer2/Target/Darwin/Thread.h
```diff
@@ -40,11 +40,12 @@ class Thread : public ds2::Target::POSIX::Thread {
   void updateState() override;
 
 public:
-  virtual ErrorCode step(int signal,
+  virtual ErrorCode suspend() override;
```
sas: `override` implies `virtual`. You shouldn't put both.

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,7 +33,50 @@ namespace Darwin {
 using Host::Darwin::LibProc;
 
 Thread::Thread(ds2::Target::Process *process, ThreadId tid)
-    : super(process, tid), _lastSyscallNumber(-1) {}
+    : super(process, tid), _lastSyscallNumber(-1) {
+  //
+  // Initially the thread is stopped.
```
sas: This comment is useless, you should remove it.

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,7 +33,50 @@ namespace Darwin {
...
+  //
+}
+
+ErrorCode Thread::suspend() {
+  ErrorCode error = kSuccess;
```
sas: I think you can simplify this function a lot by doing something like `if (_state == kRunning) return proces()->...suspend(...);` and then
```
if (_state == kTerminated) {
    return kErrorProcessNotFound;
} 

return kSuccess;
```

Sources/Target/Darwin/Thread.cpp
```diff
@@ -33,7 +33,50 @@ namespace Darwin {
...
+ErrorCode Thread::resume(int signal, Address const &address) {
+  if (_state == kTerminated) {
+    return kErrorProcessNotFound;
+  }
+
+  if (_state == kInvalid || _state == kRunning) {
+    return kErrorInvalidArgument;
+  }
+
+  signal = 0; // _stopInfo.signal;
```
sas: Why do you set signal to 0? You should add a comment to explain this, but I suppose it's because we can't send signals to processes on the Darwin implementation yet. If that's the case, you should just `DS2ASSERT(signal != 0)` at the beginning and mark it as TODO.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -219,7 +219,58 @@ ErrorCode Mach::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
...
+  unsigned int info_count = THREAD_BASIC_INFO_COUNT;
+
+  if (thread == THREAD_NULL) {
+    return kErrorProcessNotFound;
+  }
+
+  kret = mach_msg(&_reply.hdr, MACH_SEND_MSG | MACH_SEND_TIMEOUT,
+                  _reply.hdr.msgh_size, 0, MACH_PORT_NULL, 100, MACH_PORT_NULL);
+  if (kret != KERN_SUCCESS) {
+    DS2LOG(Error, "Failed to reply");
```
sas: Not very useful.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -219,7 +219,58 @@ ErrorCode Mach::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
...
+    return kErrorUnknown;
```
sas: Shouldn't you use `Platform::TranslateKernError()` for these?

Sources/Host/Darwin/Mach.cpp
```diff
@@ -219,7 +219,58 @@ ErrorCode Mach::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
...
+  kret = thread_info((thread_t)thread, THREAD_BASIC_INFO, (thread_info_t)&info,
+                     &info_count);
+  if (kret != KERN_SUCCESS) {
+    return kErrorProcessNotFound;
+  }
+  DS2ASSERT(info.suspend_count == 0);
+
+  struct task_basic_info tinfo;
+  mach_msg_type_number_t count = TASK_BASIC_INFO_COUNT;
+  kret = ::task_info(task, TASK_BASIC_INFO, (task_info_t)&tinfo, &count);
```
sas: it's good that you explicitly reference the root namespace with `::task_info`. You should do the same for other system functions.

Sources/Host/Darwin/Mach.cpp
```diff
@@ -219,7 +219,58 @@ ErrorCode Mach::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
...
+  if (kret != KERN_SUCCESS) {
+    return Platform::TranslateKernError(kret);
+  }
+
+  if (tinfo.suspend_count != 0) {
+    DS2LOG(Debug, "Resume task");
```
sas: Not useful.

Headers/DebugServer2/Target/Darwin/Thread.h
kokarez: 👍

Sources/Host/Darwin/Mach.cpp
kokarez: Yep both. I've miss these line when I've clean.

sas:
This PR is outdated. Closing for now, might get back to it when we work on macOS support more seriously in the future.

PR #199, opened 2016-02-22
Only disable RegisterVariables tests on gcc

fjricci:


a20012251:
so funny that you github recommends you to be your own reviewer

fjricci:
Yeah, I'm running the suite on lldb master to make sure the failure condition is still true.

PR #200, opened 2016-02-23
Fix watchpoint query packet format

fjricci:


fjricci:
Committed as cafe3d5d52a0ed55df7f2b773efd0f56f963fb75

sas:
Nice. Looks like this code had never been tested :)

PR #203, opened 2016-02-25
Use tid instead of pid for ptrace calls.

fjricci:
ptrace calls will fail if the main thread is sleeping at the time of the call


sas:
@facebook-github-bot shipit

PR #210, opened 2016-03-04
Implementation of QEnableCompression

kokarez:
This feature is supported by lldb from 3.7.0.

I think it will be test on Travis (will check) since it's enable by default on linux. But I didn't found specific test for it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -259,6 +259,10 @@ if (${ENABLE_LOGCAT})
   target_compile_definitions(ds2 PRIVATE ENABLE_LOGCAT)
 endif ()
 
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
+  target_link_libraries(ds2 z)
+endif ()
+
 if ("${CMAKE_SYSTEM_NAME}" MATCHES "FreeBSD")
   target_link_libraries(ds2 util procstat)
 endif ()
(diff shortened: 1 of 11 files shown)


fjricci:
I wonder if this is the error you're seeing: https://bugs.launchpad.net/ubuntu/+source/zlib/+bug/1155307

kokarez:
Yes it is :(

Here is what I've tried:
  - The _ln_ from the bug -> file not found
  - I did a find / -name zconf.h (nothing)
  - I try to add zlib1g-dev:i386 but travis don't have this package (?)
  - The last try was with lib32z1-dev, but didn't fix the issue

I'm short on idea now :)

fjricci:
Did you add zlib1g-dev:i386 using .travis.yml or apt-get install in the install.py script (you should do the latter)?

The before-install.py script will let you add a new package source if this package isn't available with what's currently on travis.

kokarez:
I used install.py. I just tried dpkg to add i386 arch also and didn't
works. Adding an other source could not work since the pakage itself exist
but is buggy.

Every command I run on travis, I run it on my local setup too, but that
doesn't help :)

So I can add the zconf.h in the repos (I've see one project with that) or
found a reliable package source. I continue to work on it :)
Le 11 mars 2016 1:01 PM, "Francis Ricci" <notifications@github.com> a
écrit :

sas:
This branch can't go in in its current state (breaks the build on Darwin, doesn't work on travis with Ubuntu, and there are a few design issues). I'm just going to close this, we can re-open in the future if we deem this useful.

PR #211, opened 2016-03-08
[windows][arm] Support compilation for windows ARM

a20012251:


sas:
This is awesome. I put a few comments on your commits, you should fix them, as well as run clang-format on your code before we can merge this.

Also, we don't do the `[windows][arm]` kind of stuff in this project, you should probably just have a proper commit message without any tagging of this sort.

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+  return 0;
+#endif
+}
+
+template <typename T> static inline int PopCount(T number) {
+#if __has_builtin(__builtin_popcount)
+  return __builtin_popcount(number);
+#else
```
kastiglione: Can you use `__popcnt` and co? https://msdn.microsoft.com/en-us/library/bb385231(v=vs.100).aspx)

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,47 @@
...
+#include "DebugServer2/Utils/CompilerSupport.h"
+
+namespace ds2 {
+namespace Utils {
+
+template <typename T> static inline int FFS(T number) {
+#if __has_builtin(__builtin_ffs)
+  return __builtin_ffs(number);
+#else
+  for (int i = 1; number != 0; i++, number >>= 1) {
```
kastiglione: Can this use `_BitScanForward`? https://msdn.microsoft.com/en-us/library/wfd9z0bb(v=vs.80).aspx

Headers/DebugServer2/Utils/Bits.h
a20012251: __popcnt and friends are not available for windows arm :(

compnerd: Is `__popcnt` available in MSVC and not in clang/clang-cl?  Or do you mean its not available in MSVC?

a20012251: it's not part of the intrinsics for the winstore api (msvc i suppose).

compnerd: The intrinsics should not be impacted by the WINAPI FAMILY.  Seems odd that the intrinsic would be unavailable since it should map pretty easily to `clz`.

sas: He's not saying they are impacted by the WINAPI FAMILY. As far as I can
understand, they're simply not available when targetting ARM, regardless of
the FAMILY.

On Tue, Mar 8, 2016 at 12:09 PM, Saleem Abdulrasool <
notifications@github.com> wrote:

> In Headers/DebugServer2/Utils/Bits.h
> <https://github.com/facebook/ds2/pull/211#discussion_r55419512>:
>
> > +#if __has_builtin(__builtin_ffs)
> > +  return __builtin_ffs(number);
> > +#else
> > +  for (int i = 1; number != 0; i++, number >>= 1) {
> > +    if (number & 1) {
> > +      return i;
> > +    }
> > +  }
> > +  return 0;
> > +#endif
> > +}
> > +
> > +template <typename T> static inline int PopCount(T number) {
> > +#if __has_builtin(__builtin_popcount)
> > +  return __builtin_popcount(number);
> > +#else
>
> The intrinsics should not be impacted by the WINAPI FAMILY. Seems odd that
> the intrinsic would be unavailable since it should map pretty easily to
> clz.
>
> —
> <https://github.com/facebook/ds2/pull/211/files#r55419512>.
>



-- 
Stephane Sezer

a20012251: yes, exactly that

kastiglione: There's a `_CountOneBits`, can it be used? https://msdn.microsoft.com/en-us/library/hh875058.aspx

kastiglione: There's also `bitset::count`.

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,60 @@
...
+#ifndef __DebugServer2_Utils_Bits_h
+#define __DebugServer2_Utils_Bits_h
+
+#include "DebugServer2/Utils/CompilerSupport.h"
+
+namespace ds2 {
+namespace Utils {
+namespace {
+// We apply builtin bit functions on chunks of 32 bits.
+const size_t kChunkSize = 32;
```
kastiglione: `sizeof(int) * 4`?

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,60 @@
...
+template <typename T> static inline int FFS(T number) {
+#if __has_builtin(__builtin_ffs)
+  while (number != 0) {
+    int ffs = __builtin_ffs(number);
+    if (ffs != 0) {
+      return ffs;
+    }
+    number >>= kChunkSize;
+  }
+#else
```
kastiglione: Was there an answer to using `_BitScanForward`?

Headers/DebugServer2/Utils/Bits.h
a20012251: Indeed, countonebits and _BitScanForward exist. I'll try to make them work

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,60 @@
...
+}
+
+template <typename T> static inline int PopCount(T number) {
+  int count = 0;
+#if __has_builtin(__builtin_popcount)
+  while (number != 0) {
+    count += __builtin_popcount(number);
+    number >>= kChunkSize;
+  }
+#else
```
kastiglione: `_CountOneBits`?

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,82 @@
...
+  }
+  return 0;
+}
+
+static inline unsigned int PopCount(unsigned int number) {
+  int count = 0;
+#if __has_builtin(__builtin_popcount)
+  while (number != 0) {
+    count += __builtin_popcount(number);
+    number >>= kChunkSize;
```
kastiglione: This shifting isn't needed here, right?

Headers/DebugServer2/Utils/Bits.h
sas: `sizeof(int) * 8` actually.

On Tue, Mar 8, 2016 at 3:28 PM, Dave Lee <notifications@github.com> wrote:

> In Headers/DebugServer2/Utils/Bits.h
> <https://github.com/facebook/ds2/pull/211#discussion_r55447727>:
>
> > +// This source code is licensed under the University of Illinois/NCSA Open
> > +// Source License found in the LICENSE file in the root directory of this
> > +// source tree. An additional grant of patent rights can be found in the
> > +// PATENTS file in the same directory.
> > +//
> > +
> > +#ifndef __DebugServer2_Utils_Bits_h
> > +#define __DebugServer2_Utils_Bits_h
> > +
> > +#include "DebugServer2/Utils/CompilerSupport.h"
> > +
> > +namespace ds2 {
> > +namespace Utils {
> > +namespace {
> > +// We apply builtin bit functions on chunks of 32 bits.
> > +const size_t kChunkSize = 32;
>
> sizeof(int) * 4?
>
> —
> <https://github.com/facebook/ds2/pull/211/files#r55447727>.
>



-- 
Stephane Sezer

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,82 @@
...
+#ifndef __DebugServer2_Utils_Bits_h
+#define __DebugServer2_Utils_Bits_h
+
+#include "DebugServer2/Utils/CompilerSupport.h"
+
+namespace ds2 {
+namespace Utils {
+namespace {
+// We apply builtin bit functions on chunks of sizeof(int) * 4 bits.
+const size_t kChunkSize = sizeof(int) * 4;
```
kastiglione: lol I said 4 in my comment but this should be 8 :trollface:

a20012251: lol

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,82 @@
...
+  }
+#endif
+  return count;
+}
+
+template <typename T> static inline unsigned int PopCount(T number) {
+  int count = 0;
+  while (number != 0) {
+    count += PopCount(number);
+    number >>= kChunkSize;
```
kastiglione: This implementation could be:

```c++
std::bitset<sizeof(T) * 8> x(number);
return x.count();
```

sas: Are you sure `std::bitset` is as performant as the implementation @a20012251
<https://github.com/a20012251> wrote?

On Tue, Mar 8, 2016 at 4:28 PM, Dave Lee <notifications@github.com> wrote:

> In Headers/DebugServer2/Utils/Bits.h
> <https://github.com/facebook/ds2/pull/211#discussion_r55453958>:
>
> > +#elif defined(_MSC_VER)
> > +  count = _CountOneBits(number);
> > +#else
> > +  while (number != 0) {
> > +    count++;
> > +    number = number & (number - 1);
> > +  }
> > +#endif
> > +  return count;
> > +}
> > +
> > +template <typename T> static inline unsigned int PopCount(T number) {
> > +  int count = 0;
> > +  while (number != 0) {
> > +    count += PopCount(number);
> > +    number >>= kChunkSize;
>
> This implementation could be:
>
> std::bitset<sizeof(T) * 8> x(number);return x.count();
>
> —
> <https://github.com/facebook/ds2/pull/211/files#r55453958>.
>



-- 
Stephane Sezer

a20012251:
omg, i fucking hate github

Headers/DebugServer2/Utils/Bits.h
kastiglione: @sas no I don't have any benchmarks, but that goes both ways. I tend to prefer to reuse than reimplement, which is why my feedback. If we want performance, I'll mention there is a binary search implementation to do `FFS`.

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,61 @@
...
+      return ffs;
+    }
+    number >>= chunkSize;
+  }
+  return 0;
+}
+
+template <typename T> static inline unsigned int PopCount(T number) {
+  std::bitset<sizeof(T) * 8> bits(number);
+  return bits.count();
```
kastiglione: As @sas pointed out, your original implementation may be more optimized than this. Code reuse is also a good thing, using `bitset` was a suggestion with that in mind.

Headers/DebugServer2/Utils/Bits.h
a20012251: I don't think bsearch would be much faster for an average of 16 bits on, but I can implement it. Still, no benchmarks...

kastiglione: yeah I didn't suggest it for a reason :stuck_out_tongue_winking_eye:

kastiglione:
:+1:

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,88 @@
...
+#include <type_traits>
+
+#include "DebugServer2/Utils/CompilerSupport.h"
+
+namespace ds2 {
+namespace Utils {
+namespace {
+// We apply the builtin bit functions on chunks of sizeof(int) * 8 bits.
+// However, we can't do full shifts of this size because only the least
+// signifitive bits are used and clang complains, so we shift twice half
```
kastiglione: whoa, what's the compiler error? is it to do with signed shifts?

a20012251: Well, if you do
```
int x = 1;
int b = 32;
x <<= b;
```
clang complaints because b must be lower than the size of x. In this specific case, the compiler or processor would only keep the last 5 bits of b and this would be like `x <<= 0`. 
I just learned something today

kastiglione: Here's a possible implementations to make this problem go away:

```c++
template <typename T> static inline unsigned int FFS(T number) {
  if (sizeof(T) <= sizeof(kIntBits)) {
    return FFS(uNumber);
  }

  typename std::make_unsigned<T>::type uNumber = number;
  while (uNumber > 0) {
    int ffs = FFS(uNumber);
    if (ffs != 0) {
      return ffs;
    }
    uNumber >>= kIntBits;
  }
  return 0;
}
```

a20012251: same error
/Users/wallace/Source/ds2/Headers/DebugServer2/Utils/Bits.h:69:13: error: right shift count >= width of type [-Werror]
     uNumber >>= kIntBits;

it seems that clang is not smart enough

a20012251: another option is to avoid templates...

kastiglione: > it seems that clang is not smart enough

ah, yeah

> another option is to avoid templates…

yeah.

I've definitely caused more than enough thrash on this. I'm going get out of your way now. :bow:

a20012251: :(

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,88 @@
...
+#define __DebugServer2_Utils_Bits_h
+
+#include <type_traits>
+
+#include "DebugServer2/Utils/CompilerSupport.h"
+
+namespace ds2 {
+namespace Utils {
+namespace {
+// We apply the builtin bit functions on chunks of sizeof(int) * 8 bits.
```
sas: Blank line after `namespace {` plz.

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,88 @@
...
+// However, we can't do full shifts of this size because only the least
+// signifitive bits are used and clang complains, so we shift twice half
+// the size.
+const size_t kHalfIntBits = sizeof(int) * 4;
```
sas: `static const` instead of just `const`.

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,88 @@
...
+}
+
+static inline unsigned int FFS(unsigned int number) {
+#if __has_builtin(__builtin_ffs)
+  return __builtin_ffs(number);
+#elif defined(_MSC_VER)
+  unsigned long index;
+  if (_BitScanForward(&index, number)) {
```
sas: The prototype of `_BitScanForward` sounds extra dumb.

a20012251: windows ftw

Headers/DebugServer2/Utils/Bits.h
```diff
@@ -0,0 +1,88 @@
...
+// We apply the builtin bit functions on chunks of sizeof(int) * 8 bits.
+// However, we can't do full shifts of this size because only the least
+// signifitive bits are used and clang complains, so we shift twice half
+// the size.
+const size_t kHalfIntBits = sizeof(int) * 4;
+}
+
+static inline unsigned int FFS(unsigned int number) {
+#if __has_builtin(__builtin_ffs)
+  return __builtin_ffs(number);
```
sas: I don't think it matters here, but `__builtin_ffs` takes an `int` argument, not `unsigned int`.

Headers/DebugServer2/Utils/Bits.h
sas: Least significant bits, not signifitive

Headers/DebugServer2/Utils/Bits.h
a20012251: I know, but that way my manual algorithm will work. doing right shifts on negative numbers maintain the sign bit.

Headers/DebugServer2/Utils/Bits.h
sas: Because that way, `kHalfIntBits` is local to this translation unit and
might be elided at compilation. Not marking it `static` forces the compiler
to emit a symbol for it (because it can be used by other translation units).

On Thu, Mar 10, 2016 at 11:15 AM, walter erquinigo <notifications@github.com
> wrote:

> In Headers/DebugServer2/Utils/Bits.h
> <https://github.com/facebook/ds2/pull/211#discussion_r55733368>:
>
> > +
> > +#ifndef __DebugServer2_Utils_Bits_h
> > +#define __DebugServer2_Utils_Bits_h
> > +
> > +#include <type_traits>
> > +
> > +#include "DebugServer2/Utils/CompilerSupport.h"
> > +
> > +namespace ds2 {
> > +namespace Utils {
> > +namespace {
> > +// We apply the builtin bit functions on chunks of sizeof(int) * 8 bits.
> > +// However, we can't do full shifts of this size because only the least
> > +// signifitive bits are used and clang complains, so we shift twice half
> > +// the size.
> > +const size_t kHalfIntBits = sizeof(int) * 4;
>
> teach me why, please
>
> —
> <https://github.com/facebook/ds2/pull/211/files#r55733368>.
>



-- 
Stephane Sezer

Headers/DebugServer2/Utils/Bits.h
sas: Makes sense.

On Thu, Mar 10, 2016 at 11:19 AM, walter erquinigo <notifications@github.com
> wrote:

> In Headers/DebugServer2/Utils/Bits.h
> <https://github.com/facebook/ds2/pull/211#discussion_r55734004>:
>
> > +
> > +#include "DebugServer2/Utils/CompilerSupport.h"
> > +
> > +namespace ds2 {
> > +namespace Utils {
> > +namespace {
> > +// We apply the builtin bit functions on chunks of sizeof(int) * 8 bits.
> > +// However, we can't do full shifts of this size because only the least
> > +// signifitive bits are used and clang complains, so we shift twice half
> > +// the size.
> > +const size_t kHalfIntBits = sizeof(int) * 4;
> > +}
> > +
> > +static inline unsigned int FFS(unsigned int number) {
> > +#if __has_builtin(__builtin_ffs)
> > +  return __builtin_ffs(number);
>
> I know, but that way my manual algorithm will work. doing right shifts on
> negative numbers maintain the sign bit.
>
> —
> <https://github.com/facebook/ds2/pull/211/files#r55734004>.
>



-- 
Stephane Sezer

Headers/DebugServer2/Utils/Bits.h
kastiglione: > extra dumb

:trollface:

Headers/DebugServer2/Utils/Bits.h
kastiglione: According to [cppreference.com](http://en.cppreference.com/w/cpp/language/storage_duration)

> Any of the following names declared at namespace scope have internal linkage
> …
> * non-volatile const-qualified variables (including `constexpr`) that aren't declared `extern` and aren't previously declared to have external linkage.

sas: > *An unnamed namespace or a namespace declared directly or indirectly
within an unnamed namespace has internal linkage*.

I hadn't seen that `namespace {}` was enclosing this. So yeah, `static` is
useless in this case.

On Thu, Mar 10, 2016 at 11:44 AM, Dave Lee <notifications@github.com> wrote:

> In Headers/DebugServer2/Utils/Bits.h
> <https://github.com/facebook/ds2/pull/211#discussion_r55738070>:
>
> > +
> > +#ifndef __DebugServer2_Utils_Bits_h
> > +#define __DebugServer2_Utils_Bits_h
> > +
> > +#include <type_traits>
> > +
> > +#include "DebugServer2/Utils/CompilerSupport.h"
> > +
> > +namespace ds2 {
> > +namespace Utils {
> > +namespace {
> > +// We apply the builtin bit functions on chunks of sizeof(int) * 8 bits.
> > +// However, we can't do full shifts of this size because only the least
> > +// signifitive bits are used and clang complains, so we shift twice half
> > +// the size.
> > +const size_t kHalfIntBits = sizeof(int) * 4;
>
> According to cppreference.com
> <http://en.cppreference.com/w/cpp/language/storage_duration>
>
> Any of the following names declared at namespace scope have internal
> linkage
> …
>
>    - non-volatile const-qualified variables (including constexpr) that
>    aren't declared extern and aren't previously declared to have external
>    linkage.
>
> —
> <https://github.com/facebook/ds2/pull/211/files#r55738070>.
>



-- 
Stephane Sezer

PR #213, opened 2016-03-09
Add missing x86_64 lldb registers to CPUState header

fjricci:


sas:
@facebook-github-bot shipit

PR #214, opened 2016-03-11
Builds for winstore ARM 

a20012251:


sas:
Still a couple of small comments, but overall this looks like it's almost ready to go.

sas:
@facebook-github-bot shipit

PR #216, opened 2016-03-11
Add the clang-only flag -Wshorten-64-to-32

a20012251:
I mainly used these two replacement commands because I trust the compiler more than myself

for: X = Y;
.s/\( *\)\(.*\) = \(.*\);/\1\2 = static_cast<typeof(\2)>(\3);

for: type X = Y;
.s/\( *\)\(.*\) \(.*\) = \(.*\);/\1\2 \3 = static_cast<\2>(\4);

very few cases required an explicit type


kastiglione:
Are there cases where this warning will come in handy? It seems like the change adds the warning and then silences it with casts throughout.

a20012251:
Well, my intention right now is not to fix any possible bug. However, this
flag will make new code mindful of casting, hopefully preventing bugs.

2016-03-11 9:10 GMT-08:00 Dave Lee <notifications@github.com>:

> Are there cases where this warning will come in handy? It seems like the
> change adds the warning and then silences it with casts throughout.
>
> —
> <https://github.com/facebook/ds2/pull/216#issuecomment-195458985>.
>



-- 
Walter Erquínigo Pezo

Every problem has a simple, fast and wrong solution.

a20012251:
well, i'll verify each cast

sas:
Should we close this and get rid of the branch?

PR #217, opened 2016-03-11
Record full ds2 launch command to the Debug log

fjricci:
Store command at the beginning because we modify argv while parsing. Log towards the end of main, so that we're sure the desired log level has been set before logging.

### DIFF
--- a/Sources/main.cpp
+++ b/Sources/main.cpp
@@ -233,6 +233,11 @@ DS2_ATTRIBUTE_NORETURN static void ListProcesses() {
 }
 
 int main(int argc, char **argv) {
+  std::ostringstream ss;
+  for (int i = 0; i < argc; ++i)
+    ss << argv[i] << " ";
+  std::string full_args = ss.str();
+
   ds2::OptParse opts;
   int idx;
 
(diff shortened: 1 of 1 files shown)


Sources/main.cpp
```diff
@@ -233,6 +233,10 @@ DS2_ATTRIBUTE_NORETURN static void ListProcesses() {
 }
 
 int main(int argc, char **argv) {
+  std::vector<std::string> full_args;
```
sas: `std::vector<std::string>(argv[0], argv[argc]);`

Sources/main.cpp
```diff
@@ -422,6 +426,10 @@ int main(int argc, char **argv) {
     host = gDefaultHost;
   }
 
+  DS2LOG(Debug, "DS2 Launched with Arguments:");
+  for (auto const &arg : full_args)
+    DS2LOG(Debug, arg.c_str());
```
sas: `DS2LOG(Debug, "  %s", arg.c_str());`

sas:
@facebook-github-bot shipit

PR #218, opened 2016-03-11
Builds for winstore arm

a20012251:
Depends on https://github.com/facebook/ds2/pull/214


sas:
Why do you need this? I think you should be able to just build these functions without modification if you use my earlier PR from today.

a20012251:
lol, this pr was not necessary :D

PR #220, opened 2016-03-12
Minimal config for Appveyor with target x86

a20012251:
the build passed according to: https://ci.appveyor.com/project/a20012251/ds2/build/1.0.14

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,20 @@
+# Operating system (build VM template)
+os: Visual Studio 2015
+
+clone_folder: c:\projects\ds2
+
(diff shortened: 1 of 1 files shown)


sas:
This builds the x86 version anyway. You have to run cmake with the "Win64" generator to run x64 builds.

fjricci:
In that case, :+1:

sas:
You should add `clone_depth: 1` to the config file. I suppose it costs nothing, and would speed up the cloning process a little bit.

Also, you might wanna use the file I added in PR#221 for these builds.

PR #223, opened 2016-03-14
Add Appveyor 

a20012251:


sas:
The toolchain file is empty and the comment in there is not relevant for Windows (only WinPhone).

a20012251:
It works without the file, so I removed that trash.

sas:
@facebook-github-bot shipit

a20012251:
commited as 3ea38519c5a7ca0b4dbd14cd9bcabcc808941daf

PR #225, opened 2016-03-15
Make sure ds2 includes follow llvm sorting policy

fjricci:


sas:
@facebook-github-bot shipit

PR #229, opened 2016-03-15
Use understandable names for breakpoint control and address registers

fjricci:


Sources/Host/Linux/ARM/PTraceARM.cpp
```diff
@@ -115,26 +115,26 @@ ErrorCode PTrace::readCPUState(ProcessThreadId const &ptid, ProcessInfo const &,
...
 
     if (wrapPtrace(PTRACE_GETHBPREGS, pid, (n << 1) + 1, &value) < 0) {
-      state.hbp.bcr[n] = 0;
+      state.hbp.bp_addr[n] = 0;
     } else {
-      state.hbp.bcr[n] = value;
+      state.hbp.bp_addr[n] = value;
     }
 
     if (wrapPtrace(PTRACE_GETHBPREGS, pid, (n << 1) + 2, &value) < 0) {
```
sas: I wonder why we use a temp variable here (`value`) and we don't for watchpoints?

sas:
@facebook-github-bot shipit

Sources/Host/Linux/ARM/PTraceARM.cpp
fjricci: I'm currently overhauling this pretty significantly currently. So I will make it more consistent as I go :smiley:

fjricci:
Clang-formatted and committed as fe86030c2597cc4a9769ca6de0fbedc0618330e7

PR #231, opened 2016-03-16
Remove some duplicate code in PTrace.

sas:
A lot of code is identical between the Linux, FreeBSD and Darwin
implementations of our ptrace abstraction. This commit moves a couple of
functions up into the POSIX implementation.

There are probably more methods that could be moved in the same way.

### DIFF
--- a/Sources/Host/POSIX/PTrace.cpp
+++ b/Sources/Host/POSIX/PTrace.cpp
@@ -9,6 +9,7 @@
 //
 
 #include "DebugServer2/Host/POSIX/PTrace.h"
+#include "DebugServer2/Host/Platform.h"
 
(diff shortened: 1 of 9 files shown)


fjricci:
@facebook-github-bot shipit

Sources/Host/POSIX/PTrace.cpp
```diff
@@ -43,6 +50,30 @@ ErrorCode PTrace::wait(ProcessThreadId const &ptid, int *status) {
   return kSuccess;
 }
 
+ErrorCode PTrace::attach(ProcessId pid) {
+  if (pid <= kAnyProcessId)
```
a20012251: curly

PR #234, opened 2016-03-21
Implement query functions for number of supported hardware watchpoints

fjricci:


sas:
Looks like your code doesn't pass coding style, and is missing a couple of PTrace function on ARM64.

sas:
Also breaks Windows on ARM :disappointed:

Sources/Host/Linux/ARM/PTraceARM.cpp
```diff
@@ -136,6 +137,43 @@ ErrorCode PTrace::readCPUState(ProcessThreadId const &ptid, ProcessInfo const &,
...
+
+  if (((value >> 24) & 0xff) == 0) {
+    return 0;
+  }
+
+  return value;
+}
+
+int PTrace::getMaxHardwareBreakpoints(ProcessThreadId const &ptid) {
+  return std::min<int>(getStoppointData(ptid) & 0xff, ds2::kMaxHWStoppoints);
```
sas: I don't think `kMaxHWStoppoints` is defined in this commit.
Also, I'm fairly certain you don't need to specify the template type (`int`) for `std::max` because it will be deduced from arguments, unless you have a type mismatch.

fjricci: kMaxHWStoppoints is defined in this commit, in DebugServer2/HardwareBreakpointManager.h

So the type mismatch comes from comparing to an enum. But I can specify the enum type to get around that.

Sources/Architecture/ARM/HardwareBreakpointManager.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+#define super ds2::BreakpointManager
+
+namespace ds2 {
+namespace Architecture {
+namespace ARM {
+
+HardwareBreakpointManager::HardwareBreakpointManager(Target::Process *process)
+    : super(process) {
+  _maxBreakpoints = _process->getMaxBreakpoints();
```
sas: You should initialize `_maxBreakpoints` and friends in the initializer list (like `super`).

Sources/Architecture/X86/HardwareBreakpointManager.cpp
```diff
@@ -0,0 +1,31 @@
...
+namespace X86 {
+
+HardwareBreakpointManager::HardwareBreakpointManager(Target::Process *process)
+    : super(process) {}
+
+HardwareBreakpointManager::~HardwareBreakpointManager() {}
+
+int HardwareBreakpointManager::maxWatchpoints() {
+  // dr0, dr1, dr2, dr3
+  return 4;
```
sas: <3

sas:
@facebook-github-bot shipit

PR #235, opened 2016-03-21
Make ds2 build for WinPhone-ARM.

sas:


fjricci:
@facebook-github-bot shipit

Headers/DebugServer2/Host/Windows/ExtraWrappers.h
```diff
@@ -358,23 +357,80 @@ BOOL WINAPI VirtualFreeEx(
...
+//
+// If we need more functions that are only available in kernel32.dll, we just
+// need to add their prototype in the above `extern "C" {}` block, and add a
+// one line `#define` below, in this form:
+//     #define FunctionName(...) DO_K32_CALL(FunctionName, __VA_ARGS__)
+
+static inline FARPROC GetK32Proc(_In_ LPCSTR procName) {
+  static HMODULE kernel32Base = nullptr;
+  if (kernel32Base == nullptr) {
+    kernel32Base = GetModuleHandle(L"kernel32");
```
kastiglione: Is there any chance of the double initialization race here being a problem?

Headers/DebugServer2/Host/Windows/ExtraWrappers.h
```diff
@@ -358,23 +357,80 @@ BOOL WINAPI VirtualFreeEx(
...
+  }
+
+  return GetProcAddress(kernel32Base, procName);
+}
+
+template <typename RetType, typename... ArgTypes>
+RetType CallK32Proc(char const *name, ArgTypes... args) {
+  auto procAddress =
+      reinterpret_cast<RetType (*)(ArgTypes...)>(GetK32Proc(name));
+  return procAddress(args...);
```
kastiglione: :trophy:

Headers/DebugServer2/Host/Windows/ExtraWrappers.h
sas: There's only one thread in ds2 that can call this function. I think if
there was a race for the initialization, we could call `GetModuleHandle`
twice instead of just once, but there still wouldn't be a bug cause it's
just a pointer and it should be the same for every call.

Thanks for pointing it out though, I hadn't thought about it.

On Mon, Mar 21, 2016 at 2:15 PM, Dave Lee <notifications@github.com> wrote:

> In Headers/DebugServer2/Host/Windows/ExtraWrappers.h
> <https://github.com/facebook/ds2/pull/235#discussion_r56898808>:
>
> > +//     GetProcAddress(GetModuleHandle("kernel32"), funcName)
> > +// and call that.
> > +//
> > +// Given that we have declarations for these functions above, we can also
> > +// type-check these calls and make sure both the return type and the arguments
> > +// type (see `decltype` in `DO_K32_CALL`) are matching.
> > +//
> > +// If we need more functions that are only available in kernel32.dll, we just
> > +// need to add their prototype in the above `extern "C" {}` block, and add a
> > +// one line `#define` below, in this form:
> > +//     #define FunctionName(...) DO_K32_CALL(FunctionName, __VA_ARGS__)
> > +
> > +static inline FARPROC GetK32Proc(_In_ LPCSTR procName) {
> > +  static HMODULE kernel32Base = nullptr;
> > +  if (kernel32Base == nullptr) {
> > +    kernel32Base = GetModuleHandle(L"kernel32");
>
> Is there any chance of the double initialization race here being a problem?
>
> —
> You are receiving this because you modified the open/close state.
> <https://github.com/facebook/ds2/pull/235/files/817838778e518fff191ed2166a85717de8295c94#r56898808>
>



-- 
Stephane Sezer

PR #236, opened 2016-03-22
Make sure static builds work on all linux distributions.

sas:
When linking pthread, we need --whole-archive.


fjricci:
@facebook-github-bot shipit

PR #237, opened 2016-03-22
Improve detection of Android vs. Tizen vs. other Linux.

sas:


Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -173,7 +173,7 @@ EnumerateLinkMap(ELFProcess *process, Address addressToDPtr,
   // FIXME: main executable detection is disabled because this causes some
   // issues when using the mozzloader. LLDB seems to be fine with us not
   // marking any object as being the main executable anyway.
-  bool isMain = false;
+  bool isMain = true;
```
fjricci: The commit message says that this is only for android, but I don't see any platform-specific code?

Headers/DebugServer2/Base.h
```diff
@@ -11,10 +11,6 @@
 #ifndef __DebugServer2_Base_h
 #define __DebugServer2_Base_h
 
-#if defined(__APPLE__)
-#include "TargetConditionals.h"
-#endif
-
 #if defined(_WIN32)
```
kokarez: Did it compile without that ?

Headers/DebugServer2/Base.h
```diff
@@ -45,7 +41,7 @@ typedef SSIZE_T ssize_t;
 #define OS_WIN32
 #elif defined(__FreeBSD__)
 #define OS_FREEBSD
-#elif defined(__APPLE__) && defined(TARGET_OS_MAC)
```
kokarez: I added the TARGET_OS_MAC check to be sure that we will be on a specific architecture. Since mac/iphone/etc could have different implementation I was thinking it will be better to crash at the compil time if someone try to compile on an non compatible arch

```c
if __APPLE__
   if MAC

   if IOS

   else...
```

Headers/DebugServer2/Base.h
sas: Yes.

Headers/DebugServer2/Base.h
sas: Yeah but we'll have to do that when we actually target multiple apple targets. I'm almost certain that there's gonna be close to zero difference, given that the kernel is the same.

fjricci:
@facebook-github-bot shipit

PR #238, opened 2016-03-24
Fix handling of dynamic libraries on Linux targets.

sas:
This mostly affects debugging on Tizen and android when attaching to a running app.

### DIFF
--- a/Sources/Utils/Paths.cpp
+++ b/Sources/Utils/Paths.cpp
@@ -1,0 +1,35 @@
+//
+// Copyright (c) 2014-present, Facebook, Inc.
+// All rights reserved.
+//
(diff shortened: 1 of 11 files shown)


Sources/Utils/Paths.cpp
```diff
@@ -0,0 +1,35 @@
...
+std::string Basename(std::string const &path) {
+#if defined(OS_WIN32)
+  char sep = '\\';
+#else
+  char sep = '/';
+#endif
+
+  auto end = path.find_last_not_of(sep);
+  if (end == std::string::npos)
+    return "/";
```
fjricci: Windows?

fjricci:
@facebook-github-bot shipit

PR #239, opened 2016-03-25
Preserve calling conventions in CallK32Proc.

sas:
This allows us to call functions from kernel32.dll with the right
conventions, eg stdcall vs cdecl.


a20012251:
@facebook-github-bot shipit

PR #242, opened 2016-03-30
Use gnueabi as environment in target triple on tizen

fjricci:


sas:
@facebook-github-bot shipit

PR #249, opened 2016-04-05
Use override for all processes, as done with linux already

fjricci:


sas:
You should probably build with https://github.com/facebook/ds2/pull/250/commits/b9745110382fd00ab88047321d8109f195030898 to have the compiler help you a bit.

sas:
@facebook-github-bot shipit

PR #255, opened 2016-04-06
Implement QListThreadsInStopReply packet

fjricci:


sas:
I don't think we should store this information in `GDBRemote::StopInfo`. That structure stores the data and shouldn't hold information about how its users intend to print it.

sas:
@facebook-github-bot shipit

PR #256, opened 2016-04-06
Fix inconsistency between 4.9 and 4.9.x in aosp toolchain

fjricci:


sas:
@facebook-github-bot shipit

PR #261, opened 2016-04-07
Remove the status argument to `Process::wait`

sas:
This also puts a definition of `wait` in `ProcessBase` and adds
`override` on all implementations in lower classes.


fjricci:
@facebook-github-review-bot shipit

PR #262, opened 2016-04-07
Remove depth: 1 from travis config.

sas:
This causes issues with builds that start after we do two pushes in a
row. Travis `git clone`s and then does a checkout of the git hash that
needs to be tested. If we do a clone with a depth of 1, only the HEAD
hash will be available.


fjricci:
@facebook-github-bot shipit

PR #267, opened 2016-04-07
Fix Bits.h helpers for Windows.

sas:
A header was missing `intrin.h`, and it seems like `_CountOneBits` is
only available on ARM (can't find any declaration for this for x86).


a20012251:
@facebook-github-bot shipit
so sad

PR #268, opened 2016-04-08
Make sure we don't accept invalid values for breakpoint type.

sas:


fjricci:
@facebook-github-bot shipit

PR #274, opened 2016-04-13
Use upstream fix instead of disabling multithreaded tests

fjricci:


fjricci:
Might as well wait and keep this as part of the move to the 3.8 suite.

PR #275, opened 2016-04-14
Fix leaking fd's when launching inferior

fjricci:


sas:
@facebook-github-bot shipit

sas:
Hilarious patchset. But so good!

fjricci:
Committed as a43195d..d5abff4

PR #276, opened 2016-04-14
Return a slightly more precise error in POSIX::ProcessSpawner.

sas:
Also some fancy braces are included in this change.


fjricci:
@facebook-github-bot shipit

PR #277, opened 2016-04-14
Update cmake check to account for the switch to ninja

fjricci:


sas:
@facebook-github-bot shipit

PR #278, opened 2016-04-15
Add a `--fast` option to run-lldb-tests.sh.

sas:
This basically just runs the test. It doesn't try to build lldb, or
patch the tests, or anything. It's meant to be used by developers, after
the first run of run-lldb-tests.


fjricci:
@facebook-github-bot shipit

PR #282, opened 2016-04-15
Play around with PTrace.

sas:


fjricci:
@facebook-github-bot shipit

fjricci:
I see more merge conflicts 😢

fjricci:
@facebook-github-bot shipit

Sources/Host/Linux/PTrace.cpp
```diff
@@ -255,19 +255,16 @@ ErrorCode PTrace::suspend(ProcessThreadId const &ptid) {
 
 ErrorCode PTrace::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
                        int signal, Address const &address) {
+#if defined(ARCH_ARM)
+  DS2BUG("single stepping for ARM must be implemented in software");
+#endif
```
strager: What if we make this a compilation error by commenting out the entire definition on ARM?

strager:
First two lgtm. Third is dogscience.

Sources/Host/Linux/PTrace.cpp
sas: I thought about this, but then we have to alter the headers as well, and
I'd love to avoid having more `#ifdefs` than we already have. Not sure if
there's a cleaner way to do it at this point. I'll take a look though,
compile error would be better here.

On Mon, Apr 18, 2016 at 12:24 PM, strager <notifications@github.com> wrote:

> In Sources/Host/Linux/PTrace.cpp
> <https://github.com/facebook/ds2/pull/282#discussion_r60118811>:
>
> > @@ -255,19 +255,16 @@ ErrorCode PTrace::suspend(ProcessThreadId const &ptid) {
> >
> >  ErrorCode PTrace::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
> >                         int signal, Address const &address) {
> > +#if defined(ARCH_ARM)
> > +  DS2BUG("single stepping for ARM must be implemented in software");
> > +#endif
>
> What if we make this a compilation error by commenting out the entire
> definition on ARM?
>
> —
> <https://github.com/facebook/ds2/pull/282/files/8bcd2b3177be09d93ae1335b054bb7a354d0f5b4..55c8a56c6d24acd45b663fe1f5b0044bf5f10a6e#r60118811>
>



-- 
Stephane Sezer

sas: Yeah I just looked at this, and PTrace::step appears not only on linux, but also in the POSIX header, that's gonna be too many ifdefs, so I don't think it's worth the pain.

strager: I was thinking just hide the definition (to get a link error), not also the declaration (to get a compile error).

Sources/Host/Linux/PTrace.cpp
```diff
@@ -281,23 +273,12 @@ ErrorCode PTrace::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
       return error;
   }
 
-  if (wrapPtrace(PTRACE_SINGLESTEP, pid, nullptr, signal) < 0)
-    return Platform::TranslateError();
-
-  return kSuccess;
+  return super::step(ptid, pinfo, signal);
```
strager: Explicitly write `Address()` for `super::step`'s `address` parameter? At first glance, it looks like you forgot to pass the `address` to `super::step`.

Sources/Host/POSIX/PTrace.cpp
```diff
@@ -74,6 +74,54 @@ ErrorCode PTrace::detach(ProcessId pid) {
...
+
+#if defined(OS_LINUX)
+  // Handling of continuation address is performed in Linux::PTrace.
+  DS2ASSERT(!address.valid());
+  auto res = wrapPtrace(PTRACE_SINGLESTEP, pid, nullptr, signal);
+#elif defined(OS_FREEBSD) || defined(OS_DARWIN)
+  // (caddr_t)1 indicates that execution is to pick up where it left off.
+  auto res =
+      wrapPtrace(PT_STEP, pid, address.valid() ? address.value() : 1, signal);
+#endif
```
strager: #else
  #error "Unsupported platform"
  #endif

?

Sources/Host/Linux/PTrace.cpp
strager: Nevermind. Virtual functions. ;_;

Sources/Host/POSIX/PTrace.cpp
```diff
@@ -74,6 +74,54 @@ ErrorCode PTrace::detach(ProcessId pid) {
...
 
+ErrorCode PTrace::step(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
+                       int signal, Address const &address) {
+  pid_t pid;
+
+  ErrorCode error = ptidToPid(ptid, pid);
+  if (error != kSuccess)
+    return error;
+
+#if defined(OS_LINUX)
```
strager: I don't like this at all. =[

What's the point of all this subclass stuff anyway when only one implementation is actually compiled? @_@

strager:
9139ec6 and its parents lgtm, and 1610392 lgtm.

For bd2180e and 7e23b5e (PTrace::{step,resume} refactors), I think it makes more sense to push `POSIX::PTrace::doStepResume` down to the subclasses than to push the code up and `#if` on the platform like you have done.

I also think it makes more sense to delete these subclasses classes and put them all in one file. =]

btw, FreeBSD and Darwin are probably similar enough to be put into a common BSD implementation/class. That might cut down on the duplication.

(Does ds2 actually support Darwin/OS X?)

PR #283, opened 2016-04-19
Forward data from OutputDebugString to the debugger.

sas:
This commit contains a seemingly useless switch on the stop reason but
this will be used in later commits to implement automatic continue of
other types of stops, for instance thread creation, etc.

### DIFF
--- a/Sources/GDBRemote/DebugSessionImpl.cpp
+++ b/Sources/GDBRemote/DebugSessionImpl.cpp
@@ -942,28 +942,56 @@ DebugSessionImpl::onResume(Session &session,
     }
   }
 
-  //
   // If kErrorAlreadyExist is set, then a signal is already pending.
-  //
   if (error != kErrorAlreadyExist) {
-    //
-    // Wait for the next signal
-    //
-    error = _process->wait();
-    if (error != kSuccess)
-      goto ret;
+    bool keepGoing = true;
+    while (keepGoing) {
+      error = _process->wait();
+      if (error != kSuccess) {
+        goto ret;
+      }
+
+      auto thread = _process->currentThread();
+      if (thread == nullptr) {
+        break;
+      }
+
+      if (thread->stopInfo().event != StopInfo::kEventStop) {
+        break;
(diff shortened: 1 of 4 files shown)


Sources/GDBRemote/DebugSessionImpl.cpp
```diff
@@ -942,28 +942,54 @@ DebugSessionImpl::onResume(Session &session,
...
+      auto thread = _process->currentThread();
+      if (thread == nullptr) {
+        break;
+      }
+
+      if (thread->stopInfo().event != StopInfo::kEventStop) {
+        break;
+      }
+
+      switch (thread->stopInfo().reason) {
```
a20012251: this looks weird to me, why don't you just use an if inside the #if?

sas: Look at the commit message.

On Monday, April 18, 2016, walter erquinigo <notifications@github.com>
wrote:

> In Sources/GDBRemote/DebugSessionImpl.cpp
> <https://github.com/facebook/ds2/pull/283#discussion_r60162672>:
>
> > +    for (;;) {
> > +      error = _process->wait();
> > +      if (error != kSuccess) {
> > +        goto ret;
> > +      }
> > +
> > +      auto thread = _process->currentThread();
> > +      if (thread == nullptr) {
> > +        break;
> > +      }
> > +
> > +      if (thread->stopInfo().event != StopInfo::kEventStop) {
> > +        break;
> > +      }
> > +
> > +      switch (thread->stopInfo().reason) {
>
> this looks weird to me, why don't you just use an if inside the #if?
>
> —
> <https://github.com/facebook/ds2/pull/283/files/63dabce5f93f14243f6ed214d6283b97e23f291a#r60162672>
>


-- 
Stephane Sezer

a20012251: bleh, i'm stupid D:

a20012251:
@facebook-github-bot shipit

PR #284, opened 2016-04-19
Add error logging to process spawner

fjricci:


Sources/Host/POSIX/ProcessSpawner.cpp
```diff
@@ -330,6 +330,7 @@ ErrorCode ProcessSpawner::run(std::function<bool()> preExecAction) {
       startRedirectThread = true;
       if (term[0] == -1) {
         if (!open_terminal(term)) {
+          DS2LOG(Error, "failed to open terminal");
```
sas: You can use `Stringify::Errno` here too.

Sources/Host/POSIX/ProcessSpawner.cpp
```diff
@@ -342,6 +343,7 @@ ErrorCode ProcessSpawner::run(std::function<bool()> preExecAction) {
   _pid = ::fork();
   if (_pid < 0) {
     close_terminal(term);
+    DS2LOG(Error, "cannot fork an inferior process: %s", Stringify::Errno(errno));
```
sas: s/cannot fork an inferior process/failed to fork/

Sources/Host/POSIX/ProcessSpawner.cpp
```diff
@@ -404,6 +406,7 @@ ErrorCode ProcessSpawner::run(std::function<bool()> preExecAction) {
       environment.push_back(nullptr);
 
       if (!preExecAction()) {
+        DS2LOG(Error, "pre exec action failed, cannot execute inferior");
```
sas: s/, cannot execute inferior//

This will be logged higher up in the stack IIRC (if it isn't we should add a log there too).

sas:
@facebook-github-bot shipit

This is good but you should do the slight modifications I mentioned to the actual messages.

sas:
@facebook-github-bot shipit

PR #285, opened 2016-04-19
Resume the whole process when required.

sas:
We were resuming only one thread, which led to some fancy deadlocks.


a20012251:
@facebook-github-bot shipit

PR #286, opened 2016-04-19
Resume the whole process when required.

sas:


a20012251:
@facebook-github-bot shipit

PR #287, opened 2016-04-19
Make sure to clear register info even when unset

fjricci:


sas:
@facebook-github-bot shipit

sas:
Does that allow us to re-enable more tests?

fjricci:
Nope, there's still another bug in the numbering system, haven't decided if it's ds2 or lldb yet.

sas:
Committed as 36a879c0dcd5b485f65fdcee2e5820090130ecba.

PR #288, opened 2016-04-20
Fix tizen builds

sas:
Just stop using the android toolchain for tizen builds.


a20012251:
@facebook-github-bot shipit
seems legit

PR #289, opened 2016-04-20
Handle --setsid option.

sas:
This is important otherwise we might end up dying on Ctrl-C.


fjricci:
git rm Support/Testing/Patches/Disable-broken-Stub-tests.patch

strager:
Before this patch: The parent, ds2, and the server run in different process groups.

After this patch: Without -S, the parent, ds2, and the server run in the same process group. With -S, ds2 and the server run in a process group different from the parent.

Your commit message says: "This is important otherwise we might end up dying on Ctrl-C." When would ds2 or the server die on Ctrl-C before this patch? Both before and after this patch (with -S), Ctrl-C of the parent won't affect the server.

sas:
This is incorrect, before this patch, the parent (lldb/gdb) and ds2 always
ran in the same group, and the inferior ran in a different group. With this
patch, if -S is not specified, the behavior is unchanged, and if it is
specified, all three run in different groups.

You might have been misled by the `setsid()` call in `SlaveMain()`, which
is a different thing that only happens when we have platform mode enabled,
and spawn a slave worker. This is not the common use case. We use
`DebugMain()` for standard debugging sessions.

On Wed, Apr 20, 2016 at 10:40 AM, strager <notifications@github.com> wrote:

> Before this patch: The parent, ds2, and the server run in different
> process groups.
>
> After this patch: Without -S, the parent, ds2, and the server run in the
> same process group. With -S, ds2 and the server run in a process group
> different from the parent.
>
> Your commit message says: "This is important otherwise we might end up
> dying on Ctrl-C." When would ds2 or the server die on Ctrl-C before this
> patch? Both before and after this patch (with -S), Ctrl-C of the parent
> won't affect the server.
>
> —
> <https://github.com/facebook/ds2/pull/289#issuecomment-212530493>
>



-- 
Stephane Sezer

Sources/main.cpp
```diff
@@ -435,6 +425,14 @@ int main(int argc, char **argv) {
   // it.
   namedPipePath = opts.getString("named-pipe");
 
+  if (opts.getBool("setsid")) {
+#if defined(OS_WIN32)
+    opts.usageDie("--setsid not supported on Windows");
+#else
+    setsid();
```
fjricci: ::setsid()

fjricci:
@facebook-github-bot shipit

PR #291, opened 2016-04-21
Re-enable Create tests.

sas:


sas:
We had tried this already in PR#193. Not sure if this is going to work or not, but it passes on my CentOS machine, so maybe it will pass on travis now.

a20012251:
@facebook-github-bot shipit

sas:
Well. Looks like this worked.

PR #293, opened 2016-04-21
Fix argument order to fcntl.

sas:
This allows us to pass FdLeak tests with a log file.


fjricci:
oops

fjricci:
@facebook-github-bot shipit

strager:
`int`s

lgtm

PR #295, opened 2016-04-21
Re-enable RegisterVariable tests.

sas:


sas:
Works on my host but not here, abandoning this until we can actually handle them properly.

PR #299, opened 2016-04-23
Add missing WINBASEAPI to Windows declarations.

sas:
WINBASEAPI expands to `__declspec(dllimport)`. We need these if we want
to avoid having an additional import thunk generated for each external
function call.


a20012251:
@facebook-github-bot shipit
this is so good :D

PR #302, opened 2016-04-26
Use absolute paths for library names in svr4 packets when possible

fjricci:
NOTE: This may cause library duplication on Tizen in some cases without this lldb patch - http://reviews.llvm.org/D19557.


sas:
@facebook-github-bot shipit

What should we do about this patch then? Wait for the LLDB one to go in and cherry-pick it to release_38?

fjricci:
Yeah, I'll merge it to master today. I can put in some cherry-picks for the release_38 branch after that, and this would be one of them.

Also not sure how bad of a problem the duplication is? Because I assume release_37 will also have the same issue, and I don't really want to deal with cherry-picking down to that branch.

fjricci:
Committed as 3597b3b

PR #303, opened 2016-04-27
Get rid of AsyncProcessWaiter.

sas:
We don't use this anymore.


fjricci:
@facebook-github-bot shipit

fjricci:
+0, -157

I think this is my new favorite diff

PR #307, opened 2016-04-30
Fix writeCPUState on Windows.

sas:
Incorrect `memset` after the last magical macro patch.


a20012251:
@facebook-github-bot shipit

PR #310, opened 2016-05-03
Get rid of Process::isELFProcess.

sas:
Bad design :)

Also, this seemed fairly useless, and we should rework the ELFProcess
class to implement auxv enumration and library enumeration elsewhere.


fjricci:
@facebook-github-bot shipit

PR #312, opened 2016-05-03
Print the diff created by clang-format

kokarez:
I did this patch because I couldn't get my hands on a clang-format 3.8. brew propose me 3.7.1 or 3.9, even on a linux vm I have, getting the correction version is painful.

Even if you are using the same style "LLVM", all the different version of clang-format disagree on the final output. In my case, 3.9 force me to reorder some of my header, when 3.8 want the header in an other order. This is just frustrating.

So here what I'm proposing is to print the defect found in the test rather than just giving the name of the file.

<3

### DIFF
--- a/Support/Testing/Travis/script.sh
+++ b/Support/Testing/Travis/script.sh
@@ -71,6 +71,8 @@ check_dirty() {
     for f in "${dirty[@]}"; do
       echo "dirty: $f"
     done
+    echo "Here is the diff:"
+    git --no-pager diff
     exit 1
   fi
 }


Support/Testing/Travis/script.sh
```diff
@@ -71,6 +71,8 @@ check_dirty() {
     for f in "${dirty[@]}"; do
       echo "dirty: $f"
     done
+    echo "Here is the diff:"
+    git diff | cat
```
sas: `git --no-pager diff`

Support/Testing/Travis/script.sh
```diff
@@ -71,6 +71,8 @@ check_dirty() {
     for f in "${dirty[@]}"; do
       echo "dirty: $f"
     done
+    echo "Here is the diff:"
```
sas: You can remove this log message I think.

sas:
I like the idea. Just a couple of small nits to fix and I'll check it in after.

fjricci:
👍

Support/Testing/Travis/script.sh
kokarez: :D yea, didn't found it when I was in the man. Will do

Support/Testing/Travis/script.sh
kokarez: It just weird in the final output to not have an intro msg. As you want.

Support/Testing/Travis/script.sh
sas: Yeah it's in `man git` instead of `man git-diff`. That's probably why.

On Tue, May 3, 2016 at 12:05 PM, kokarez <notifications@github.com> wrote:

> In Support/Testing/Travis/script.sh
> <https://github.com/facebook/ds2/pull/312#discussion_r61937488>:
>
> > @@ -71,6 +71,8 @@ check_dirty() {
> >      for f in "${dirty[@]}"; do
> >        echo "dirty: $f"
> >      done
> > +    echo "Here is the diff:"
> > +    git diff | cat
>
> :D yea, didn't found it when I was in the man. Will do
>
> —
> <https://github.com/facebook/ds2/pull/312/files/e61e88015be046eac9a4a8c5946fe65bb1f9960b#r61937488>
>



-- 
Stephane Sezer

Support/Testing/Travis/script.sh
sas: Spoke with @kokarez. Will remove this before merging the patch.

sas:
Committed as 5a73c473e4ce4d940c306af2307d0c470ae74627.

PR #314, opened 2016-05-04
Fix compilation error on freebsd build

kokarez:
freebsd 10.1 have clang 3.4.1 by default and this doesn't sound to manage the flag inconsistent-missing-override. Since freebsd is compile with clang, not sure we want to force the simple user to upgrade clang version :-D

The other are just simple compilation fixes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -285,7 +285,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang"
   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND NOT "${CMAKE_CXX_COMPILER_VERSION}" VERSION_LESS "5.1")
     target_compile_options(ds2 PRIVATE -Wsuggest-override)
   endif ()
-  if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND NOT "${CMAKE_CXX_COMPILER_VERSION}" VERSION_LESS "3.5")
     target_compile_options(ds2 PRIVATE -Winconsistent-missing-override)
   endif ()
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
--- a/Sources/Host/FreeBSD/ProcStat.cpp
+++ b/Sources/Host/FreeBSD/ProcStat.cpp
@@ -8,14 +8,16 @@
 // PATENTS file in the same directory.
 //
 
-#include <libprocstat.h>
-#include <libutil.h>
 #include <string>
 #include <sys/elf.h>
 #include <sys/param.h>
 #include <sys/sysctl.h>
 #include <sys/types.h>
 #include <sys/user.h>
+// clang-format off
+#include <libprocstat.h>
+#include <libutil.h>
+// clang-format on
 
 #include "DebugServer2/Host/FreeBSD/ProcStat.h"
 #include "DebugServer2/Support/POSIX/ELFSupport.h"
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -286,7 +286,9 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang"
     target_compile_options(ds2 PRIVATE -Wsuggest-override)
   endif ()
   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-    target_compile_options(ds2 PRIVATE -Winconsistent-missing-override)
+    if (NOT "${OS_NAME}" MATCHES "FreeBSD")
```
sas: Instead of checking for the OS, you should check for the compiler version. We already do something similar for checking the gcc version, see line 285.

Sources/Target/FreeBSD/Process.cpp
```diff
@@ -201,8 +201,7 @@ ErrorCode Process::wait() {
 
   switch (_currentThread->_stopInfo.event) {
   case StopInfo::kEventNone:
-    switch (_currentThread->_stopInfo.reason) {
-    case StopInfo::kReasonNone:
+    if (_currentThread->_stopInfo.reason == StopInfo::kReasonNone) {
```
sas: This code makes no sense (the previous version). I'm not sure why we're checking `_stopInfo.reason`, so I think you can just get rid of the `if` block here. Look at the linux version of `Process:wait` and what we do when the stop event is `kEventNone`.

CMakeLists.txt
kokarez: true. but I don't know if it's specific to freebsd (incomplete port of clang ?) or due to a specific version. I did a quick search but didn't found any page with feature/version. Will search again.

Sources/Target/FreeBSD/Process.cpp
kokarez: Ahah :D I don't know if it make sense. But I can't really test, and we don't have tests, so I try to fix compilation with changing any behavior. If it's ok I will prefer to keep it like that until someone review/clean the freebsd code.

Sources/Host/FreeBSD/ProcStat.cpp
```diff
@@ -8,15 +8,16 @@
...
-#include <string>
 #include <sys/elf.h>
 #include <sys/param.h>
 #include <sys/sysctl.h>
 #include <sys/types.h>
 #include <sys/user.h>
 
+#include <libprocstat.h>
+#include <libutil.h>
+#include <string>
```
sas: I understand the problem you're trying to solve but I'm not sure this is the best way to do it. According to https://www.freebsd.org/cgi/man.cgi?query=libprocstat&sektion=3&apropos=0&manpath=FreeBSD%2010.0-RELEASE `sys/param.h` and `sys/queue.h` are required for `libprocstat.h` and only `sys/param.h` is required for `libutil.h`.

What you should do instead is to sort all the headers properly the way they were before, and then pull up `sys/param.h` and `sys/queue.h` before `libprocstat.h` and `libutil.h` and enclose these headers with `// clang-format off` and `// clang-format on` to make sure they dont' get automatically reordered in the future.

So you'd end up with something like this:
```
// clang-format off
#include <sys/param.h>
#include <sys/queue.h>
#include <libprocstat.h>
#include <libutil.h>
// clang-format on
#include <string>
#include <sys/elf.h>
#include <sys/param.h>
#include <sys/sysctl.h>
#include <sys/types.h>
#include <sys/user.h>
```

Sources/Target/FreeBSD/Process.cpp
sas: Yeah I guess if we can't test it it makes sense to keep it like that. Works for me.
I suppose the errors you were dealing with were for incomplete switch?

sas:
Merged the first two commits of this PR.

sas:
Superseded by PR#318.

PR #318, opened 2016-05-08
Fix FreeBSD build.

sas:
Some header ordering issue.


fjricci:
@facebook-github-bot shipit

PR #319, opened 2016-05-08
Add MacOSX build on travis

kokarez:


Support/Testing/Travis/after_success.sh
```diff
@@ -11,6 +11,10 @@
 
 set -eu
 
+if [[ "${TARGET-}" = "Darwin-X86_64" ]]; then
```
sas: Doesn't really matter here, but you could do `if [[ "${TARGET-}" = "Darwin-* ]] ...`.

Support/Testing/Travis/install.py
```diff
@@ -75,7 +78,10 @@
     packages.append('realpath')
 
 if len(packages) > 0:
-    if "CentOS Linux" in platform.linux_distribution():
+    if 'Darwin' in target:
+        check_call('brew install "%s" || true' % '" "'.join(packages), shell=True)
```
sas: Comment here 😆

sas:
Looks like Darwin builds end up with these warnings during the build step:
> grep: /proc/cpuinfo: No such file or directory
> /Users/travis/build/facebook/ds2/Support/Scripts/common.sh: line 45: * 5 / 4: syntax error: operand expected (error token is "* 5 / 4")

sas:
@facebook-github-bot shipit

PR #328, opened 2016-05-09
Use kFlagNewProcess instead of 0

kokarez:


sas:
Why is this required?

kokarez:
It's not a requirement, but more a question of readibility, 0 mean nothing. They are two call of initialize:

```c++
error = process->initialize(pid, kFlagAttachedProcess);
```
and
```c++
error = process->initialize(pid, 0);
```

So for my point of you, this is require ;D

sas:
Hey thanks for this. I didn't get what the problem was initially but then realized there were a couple more things that needed a bit of rework there. I cleanup up that, and other things, in PR#329.

PR #339, opened 2016-05-11
Step implementation on MacOSX

kokarez:


sas:
These two commits should probably go in as one (the first commit by itself can't work).

Also, I'm thinking leaking the implementation of stepping from Thread into Process (by having it call disableStep) might not be a great idea. Maybe we want to add an afterResume method to Thread?

kokarez:
I was also thinking about the after resume in the Thread class. But for the moment we don't need it since we don't process any internal structure from the process level. Also, we can have only one thread breaking at a time, so I thinks for the moment it's ok, and later on we could create the logic at the thread level.

sas:
Committed the first patch as 941d37155b48ce5626bada763af002a568bf15e3.

Sources/Target/Darwin/Process.cpp
```diff
@@ -330,6 +330,23 @@ ErrorCode Process::writeMemory(Address const &address, void const *data,
   return mach().writeMemory(_currentThread->tid(), address, data, length,
                             count);
 }
+
+ErrorCode Process::afterResume() {
+
```
sas: No need for this blank line.

Sources/Target/Darwin/Process.cpp
```diff
@@ -330,6 +330,23 @@ ErrorCode Process::writeMemory(Address const &address, void const *data,
...
+  // We are calling the Thread's afterResume from here, but it might make sense
+  // to have this known by ThreadBase and call it from Process::resume
```
sas: s/Process::resume/ProcessBase::resume/

Sources/Target/Darwin/X86_64/ThreadX86_64.cpp
```diff
@@ -29,6 +29,16 @@ ErrorCode Thread::step(int signal, Address const &address) {
...
 }
+
+ErrorCode Thread::afterResume() {
+  ErrorCode error = modifyRegisters([](Architecture::CPUState &state) {
+    state.state64.gp.eflags &= ~(1 << 8);
+  });
+  if (error != kSuccess) {
+    return error;
+  }
+  return kSuccess;
```
kokarez: Wait this check is useless!

sas: Indeed, the new version is better :)

On Wed, May 11, 2016 at 3:05 PM, kokarez <notifications@github.com> wrote:

> In Sources/Target/Darwin/X86_64/ThreadX86_64.cpp
> <https://github.com/facebook/ds2/pull/339#discussion_r62935238>:
>
> > @@ -29,6 +29,16 @@ ErrorCode Thread::step(int signal, Address const &address) {
> >    }
> >    return resume(signal, address);
> >  }
> > +
> > +ErrorCode Thread::afterResume() {
> > +  ErrorCode error = modifyRegisters([](Architecture::CPUState &state) {
> > +    state.state64.gp.eflags &= ~(1 << 8);
> > +  });
> > +  if (error != kSuccess) {
> > +    return error;
> > +  }
> > +  return kSuccess;
>
> Wait this check is useless!
>
> —
> <https://github.com/facebook/ds2/pull/339/files/c03f57caaf4695f786449526fd3da0b687ba2487#r62935238>
>



-- 
Stephane Sezer

sas:
@facebook-github-bot shipit

PR #340, opened 2016-05-11
Create TranslateKernError for Darwin kernel error

kokarez:
I just add the likeliest errors, but we will need to add more in the future.

### DIFF
--- a/Sources/Host/Darwin/Platform.cpp
+++ b/Sources/Host/Darwin/Platform.cpp
@@ -12,15 +12,15 @@
 #include "DebugServer2/Host/Darwin/LibProc.h"
 #include "DebugServer2/Utils/Log.h"
 
-#include <sys/utsname.h>
-
 #include <algorithm>
 #include <cctype>
 #include <cstdlib>
 #include <cstring>
 #include <fstream>
 #include <iostream>
+#include <mach/mach.h>
 #include <string>
+#include <sys/utsname.h>
 
 namespace ds2 {
 namespace Host {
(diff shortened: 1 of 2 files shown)


Sources/Host/Darwin/Platform.cpp
```diff
@@ -118,6 +118,26 @@ std::string Platform::GetThreadName(ProcessId pid, ThreadId tid) {
...
+ErrorCode Platform::TranslateKernError(kern_return_t kret) {
+  switch (kret) {
+  case KERN_SUCCESS:
+    return ds2::kSuccess;
+  case KERN_FAILURE:
+    return ds2::kErrorUnknown;
+  case KERN_MEMORY_FAILURE:
+  case KERN_INVALID_ADDRESS:
+    return ds2::kErrorInvalidAddress;
+  case KERN_PROTECTION_FAILURE:
```
sas: `KERN_PROTECTION_FAILURE` should return `kErrorInvalidAddress`. `kErrorInvalidAddress` is used for everything that could be a segfault type failure, and `KERN_PROTECTION_FAILURE` seems to be one too.

See http://unix.superglobalmegacorp.com/xnu/newsrc/osfmk/mach/kern_return.h.html

Sources/Host/Darwin/Platform.cpp
```diff
@@ -118,6 +118,26 @@ std::string Platform::GetThreadName(ProcessId pid, ThreadId tid) {
...
+  case KERN_NO_ACCESS:
+    return ds2::kErrorNoPermission;
+  case KERN_MEMORY_ERROR:
```
sas: Same for `KERN_MEMORY_ERROR`. `kErrorNoMemory` is only good for `KERN_RESOUCE_SHORTAGE` I think.

Sources/Host/Darwin/Platform.cpp
```diff
@@ -118,6 +118,26 @@ std::string Platform::GetThreadName(ProcessId pid, ThreadId tid) {
...
+  case KERN_RESOURCE_SHORTAGE:
+    return ds2::kErrorNoMemory;
+  default:
+    DS2BUG("unknown error code: %s [%d]", mach_error_string(kret), kret);
```
sas: nit: can you do `"unknown error code: %d [%s]"` instead (and swap the two following arguments). In other similar `DS2BUG`s we only print the numeric value and I'll add the string version that the system returns to match the code you just wrote here.

Sources/Host/Darwin/Platform.cpp
kokarez: Interesting, I was thinking kErrorInvalidAddress was really link a BAD_PARAM where the address is just wrong.

sas:
I like this! 👍 

There are a couple of tiny changes that you should probably do before we check this in, I'll merge it after.

Sources/Host/Darwin/Platform.cpp
kokarez: :+1:

sas:
@facebook-github-bot shipit

PR #342, opened 2016-05-12
Attempt to reduce some flaky stuff.

sas:


.travis.yml
```diff
@@ -69,7 +69,8 @@ matrix:
       osx_image: xcode7.3
       env: TARGET="Darwin-X86_64" CLANG="1"
 
-before_install: ./Support/Testing/Travis/before-install.py
-install:        ./Support/Testing/Travis/install.py
+# Use travis_retry during install to reduce the impact of network failures.
+before_install: travis_retry ./Support/Testing/Travis/before-install.py
```
kokarez: Did you try on you setup to run twice before-install/install ? Since X% of these script will be already done, perhaps some command could fail, isn't it ?

Support/Scripts/run-lldb-tests.sh
```diff
@@ -130,7 +130,7 @@ else
   export LLDB_DEBUGSERVER_PATH="$build_dir/ds2"
 fi
 
-export LLDB_TEST_TIMEOUT=45m
+export LLDB_TEST_TIMEOUT=8m
```
kokarez: From the issue: "TestLldbGdbServer was flaky with the default 10 minute timeout"

That's kind of funny, since this test failed recently :) Perhaps the timeout was due to the random failure, and not to a random runtime.

.travis.yml
sas: I think these scripts do work when called more than once in a row, but you're right, let me try.

Support/Scripts/run-lldb-tests.sh
sas: So it seems like an 8m timeout is enough. All the builds passed, and only one timed out, and it's one that's known to cause actual issues.

I'll re-run this a few times to make sure we don't make things worse with this timeout, but I think we should be good.

.travis.yml
sas: Yeah I checked and everything works fine.

fjricci:
@facebook-github-bot shipit

sas:
A second run seems to have worked on the first try.

sas:
Third run, `TestConcurrentEvents.py` timed out, which is a known issue (so not caused by the reduction in timeout length.

PR #344, opened 2016-05-12
Real handling of OutputDebugStrings on Windows.

sas:


Sources/Target/Windows/Thread.cpp
```diff
@@ -245,11 +245,26 @@ void Thread::updateState(DEBUG_EVENT const &de) {
     _stopInfo.reason = StopInfo::kReasonThreadExit;
     break;
 
-  case OUTPUT_DEBUG_STRING_EVENT:
+  case OUTPUT_DEBUG_STRING_EVENT: {
+    auto const &DSInfo = de.u.DebugString;
```
fjricci: dsInfo is probably more consistent camel-case?

sas: Mhmm, I think you're right. I'll fix it before checking in if travis passes
(don't want to get pulled down the queue again).

On Thu, May 12, 2016 at 5:29 PM, Francis Ricci <notifications@github.com>
wrote:

> In Sources/Target/Windows/Thread.cpp
> <https://github.com/facebook/ds2/pull/344#discussion_r63118655>:
>
> > @@ -245,11 +245,26 @@ void Thread::updateState(DEBUG_EVENT const &de) {
> >      _stopInfo.reason = StopInfo::kReasonThreadExit;
> >      break;
> >
> > -  case OUTPUT_DEBUG_STRING_EVENT:
> > +  case OUTPUT_DEBUG_STRING_EVENT: {
> > +    auto const &DSInfo = de.u.DebugString;
>
> dsInfo is probably more consistent camel-case?
>
> —
> <https://github.com/facebook/ds2/pull/344/files/faab52bb452d770e082f2574350e5fa9f51ef3d6#r63118655>
>



-- 
Stephane Sezer

Sources/Target/Windows/Thread.cpp
```diff
@@ -245,11 +245,26 @@ void Thread::updateState(DEBUG_EVENT const &de) {
...
+    // TODO: We need to use WaitForDebugEventEx to get unicode strings.
+    DS2ASSERT(dsInfo.fUnicode == 0);
+
+    _stopInfo.debugString.resize(dsInfo.nDebugStringLength - 1);
+    ErrorCode error = process()->readMemory(
+        reinterpret_cast<uint64_t>(dsInfo.lpDebugStringData),
+        const_cast<char *>(_stopInfo.debugString.c_str()),
+        dsInfo.nDebugStringLength - 1);
+    if (error != kSuccess) {
+      return;
```
kokarez: Perhaps you should set the state kStopped/kEventStop etc... before doing this call because in case of failure to get the string, the process will be still seen as running, isn't it ?

Perhaps you should set a default string like:  "Sorry we are unable to get the debug string"

sas: Mhmmm, that makes sense. Wanna write a patch? 😆

kokarez: I didn't see this was already merged! No, I have no way to test, so have fun :)

PR #346, opened 2016-05-19
Make remote debug argument match gdbserver

fjricci:
May as well be consistent.


sas:
@facebook-github-bot shipit

PR #349, opened 2016-05-21
Add missing spaces in gdb-remote packets

fjricci:


fjricci:
Looks like this breaks TestInlineStepping without http://reviews.llvm.org/D20509

sas:
The interesting part is that it works for x86 runs.

On Sun, May 22, 2016 at 12:39 PM, Francis Ricci <notifications@github.com>
wrote:

> Looks like this breaks TestInlineStepping without
> http://reviews.llvm.org/D20509
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/facebook/ds2/pull/349#issuecomment-220851618>
>



-- 
Stephane Sezer

fjricci:
So the other interesting part is that qfThreadInfo is only used (by lldb) as a fallback packet when qProcessInfo fails and then qC also fails.

fjricci:
Looks like it's also used as a fallback when it doesn't have a stop reply packet with threads in it. Looks like that's the case we're hitting here.

Headers/DebugServer2/GDBRemote/SessionBase.h
```diff
@@ -30,13 +30,14 @@ class SessionBase : public ProtocolHandler {
...
+  CompatibilityMode _compatMode;
 
 public:
-  SessionBase();
+  SessionBase(CompatibilityMode mode);
   virtual ~SessionBase();
 
 public:
-  virtual CompatibilityMode mode() const = 0;
+  CompatibilityMode mode() const {return _compatMode; };
```
sas: Hon. Style.

fjricci: right.

Sources/GDBRemote/Session.cpp
```diff
@@ -34,7 +34,7 @@ namespace ds2 {
 namespace GDBRemote {
 
 Session::Session(CompatibilityMode mode)
-    : _compatMode(mode), _threadsInStopReply(false) {
+    : SessionBase(mode), _threadsInStopReply(false) {
```
sas: I'm surprised we weren't calling the `SessionBase()` constructor explicitly before. This is good.

Sources/GDBRemote/SessionBase.cpp
```diff
@@ -139,6 +139,12 @@ bool SessionBase::sendACK() { return _channel->send("+", 1) == 1; }
 
 bool SessionBase::sendNAK() { return _channel->send("-", 1) == 1; }
 
+void SessionBase::addPacketSeparator(std::ostringstream &ss) {
```
sas: You should have a comment in this function to explain what we're trying to achieve.

Sources/GDBRemote/SessionBase.cpp
```diff
@@ -158,12 +158,12 @@ bool SessionBase::sendError(ErrorCode code) {
...
-  char message[4];
-  message[0] = 'E';
-  message[1] = NibbleToHex(code >> 4);
-  message[2] = NibbleToHex(code & 15);
-  message[3] = '\0';
-  return send(message);
+  std::ostringstream ss;
+  ss << "E";
+  addPacketSeparator(ss);
+  ss << std::hex << code;
```
sas: This is not good, you're changing the packet format. `std::hex` doesn't add `0` at the beginning to always have a two character-long error code.

sas:
@facebook-github-bot shipit

Sources/GDBRemote/SessionBase.cpp
```diff
@@ -139,6 +139,25 @@ bool SessionBase::sendACK() { return _channel->send("+", 1) == 1; }
...
+  switch (_compatMode) {
+  case kCompatibilityModeGDB:
+  case kCompatibilityModeGDBMultiprocess:
+    return " ";
+
+  case kCompatibilityModeLLDB:
+    return "";
+
+  case kCompatibilityModeLLDBThread:
+    DS2BUG("LLDBThreads is an invalid compatibility mode for SessionBase.");
```
sas: No ending period !@#$ 😆

PR #353, opened 2016-05-25
Implementation and fixes for some vFile commands

fjricci:


Sources/Host/POSIX/Platform.cpp
```diff
@@ -151,6 +151,16 @@ int Platform::OpenFile(std::string const &path, uint32_t flags, uint32_t mode) {
 
 bool Platform::CloseFile(int fd) { return (::close(fd) == 0); }
 
+ssize_t Platform::FileRead(int fd, std::string &buf, size_t count, size_t offset) {
+  buf.resize(count);
+
+  ssize_t nread = ::pread(fd, &buf[0], count, offset);
```
sas: Why do you use `pread` and not simply `read` here?

sas: Oh cause there's the `offset` argument and you don't want to change the offset in the file for multiple file reads I suppose.

sas:
@facebook-github-bot shipit

Sources/GDBRemote/Session.cpp
```diff
@@ -3079,12 +3079,12 @@ void Session::Handle_vFile(ProtocolInterpreter::Handler const &,
       sendError(kErrorInvalidArgument);
       return;
     }
-    uint64_t count = strtoull(eptr, &eptr, 16);
+    size_t count = strtoul(eptr, &eptr, 16);
```
kokarez: std::strtoul ?

Sources/GDBRemote/Session.cpp
```diff
@@ -3079,12 +3079,12 @@ void Session::Handle_vFile(ProtocolInterpreter::Handler const &,
...
     if (*eptr++ != ',') {
       sendError(kErrorInvalidArgument);
       return;
     }
-    uint64_t offset = strtoull(eptr, &eptr, 16);
+    size_t offset = strtoul(eptr, &eptr, 16);
```
kokarez: same

Sources/GDBRemote/DummySessionDelegateImpl.cpp
```diff
@@ -431,13 +431,25 @@ ErrorCode DummySessionDelegateImpl::onFileCreateDirectory(Session &,
...
-ErrorCode DummySessionDelegateImpl::onFileOpen(Session &, std::string const &,
-                                               uint32_t, uint32_t, int &) {
-  return kErrorUnsupported;
-}
-
-ErrorCode DummySessionDelegateImpl::onFileClose(Session &, int) {
-  return kErrorUnsupported;
+ErrorCode DummySessionDelegateImpl::onFileOpen(Session &, std::string const &path,
+                                          uint32_t flags, uint32_t mode,
+                                          int &fd) {
```
kokarez: format

Sources/Host/POSIX/FileIO.cpp
```diff
@@ -37,6 +38,29 @@ ErrorCode FileIO::CloseFile(int fd) {
...
+  if (offset > std::numeric_limits<off_t>::max()) {
+    return kErrorInvalidArgument;
+  }
+
+  off_t offArg = static_cast<off_t>(offset);
+  size_t countArg = static_cast<size_t>(count);
+
+  buf.resize(countArg);
+
+  ssize_t nRead = ::pread(fd, &buf[0], countArg, offArg);
```
kokarez: why not buf->c_str() ?

sas: buf->c_str() is `const` so the compiler will complain.

That being said, using `std::string` for this is kind of crap I think.
`U8Vector` (which should be renamed `ByteVector`) would be more
appropriate, but I know we're not the best at that elsewhere in the project.

On Wed, May 25, 2016 at 3:04 PM, kokarez <notifications@github.com> wrote:

> In Sources/Host/POSIX/FileIO.cpp
> <https://github.com/facebook/ds2/pull/353#discussion_r64661704>:
>
> > @@ -37,6 +38,29 @@ ErrorCode FileIO::CloseFile(int fd) {
> >    return kSuccess;
> >  }
> >
> > +ErrorCode FileIO::FileRead(int fd, std::string &buf, uint64_t &count,
> > +                           uint64_t offset) {
> > +  if (offset > std::numeric_limits<off_t>::max()) {
> > +    return kErrorInvalidArgument;
> > +  }
> > +
> > +  off_t offArg = static_cast<off_t>(offset);
> > +  size_t countArg = static_cast<size_t>(count);
> > +
> > +  buf.resize(countArg);
> > +
> > +  ssize_t nRead = ::pread(fd, &buf[0], countArg, offArg);
>
> why not buf->c_str() ?
>
> —
> <https://github.com/facebook/ds2/pull/353/files/00af9f8f51ac68b234e2901711792e11cf956058..56b2e8ce70a38350cc279f43e0a89254ee8209d9#r64661704>
>



-- 
Stephane Sezer

kokarez: Yea, in an other commit you use a cast to bypass that:
```c++
const_cast<char *>(_stopInfo.debugString.c_str())
```

I just point out this to have some _standart_ usage. Not sure which one is better.

sas: Yeah you're right, we aren't really consistent on this.

If we were using `U8Vector` though, we could do `buf.data()` everywhere,
without having to cast anything :)

On Wed, May 25, 2016 at 3:10 PM, kokarez <notifications@github.com> wrote:

> In Sources/Host/POSIX/FileIO.cpp
> <https://github.com/facebook/ds2/pull/353#discussion_r64662422>:
>
> > @@ -37,6 +38,29 @@ ErrorCode FileIO::CloseFile(int fd) {
> >    return kSuccess;
> >  }
> >
> > +ErrorCode FileIO::FileRead(int fd, std::string &buf, uint64_t &count,
> > +                           uint64_t offset) {
> > +  if (offset > std::numeric_limits<off_t>::max()) {
> > +    return kErrorInvalidArgument;
> > +  }
> > +
> > +  off_t offArg = static_cast<off_t>(offset);
> > +  size_t countArg = static_cast<size_t>(count);
> > +
> > +  buf.resize(countArg);
> > +
> > +  ssize_t nRead = ::pread(fd, &buf[0], countArg, offArg);
>
> Yea, in an other commit you use a cast to bypass that:
>
> const_cast<char *>(_stopInfo.debugString.c_str())
>
> I just point out this to have some *standart* usage. Not sure which one
> is better.
>
> —
> <https://github.com/facebook/ds2/pull/353/files/00af9f8f51ac68b234e2901711792e11cf956058..56b2e8ce70a38350cc279f43e0a89254ee8209d9#r64662422>
>



-- 
Stephane Sezer

fjricci: Ok yeah, I thought this was dumb. I'll see if I can move other analogous functions over to U8Vector.

sas:
Committed some of these as e4e92a36f59bedb7e5e01536553220167fd28ba4..4d136da2dff545e40f12f14f0e8b2946c6f9b550.

Headers/DebugServer2/Host/FileIO.h
```diff
@@ -0,0 +1,33 @@
...
+#include "DebugServer2/Host/Windows/FileIO.h"
+#else
+#include "DebugServer2/Host/POSIX/FileIO.h"
+#endif
+
+namespace ds2 {
+namespace Host {
+
+#if defined(OS_WIN32)
+using Windows::FileIO;
```
sas: It seems like you're keeping the interface of `FileIO` identical between POSIX and Windows (which is what we should do). If that's the case, you don't need these three header files with these inclusions; you can just put the interface of `FileIO` in `Headers/DebugServer2/Host/FileIO.h` and use it everywhere.

Headers/DebugServer2/GDBRemote/DummySessionDelegateImpl.h
```diff
@@ -12,13 +12,15 @@
...
 #include "DebugServer2/GDBRemote/SessionDelegate.h"
+#include "DebugServer2/Host/FileIO.h"
 
 namespace ds2 {
 namespace GDBRemote {
 
 class DummySessionDelegateImpl : public SessionDelegate {
 protected:
   bool _secure;
+  Host::FileIO _fileIO;
```
sas: I don't think this is good. You can have multiple `open` calls followed by multiple `close` calls. You should have an `std::list<FileIO>` or something like that.

Headers/DebugServer2/Host/Windows/FileIO.h
```diff
@@ -25,13 +25,13 @@ namespace Windows {
 
 class FileIO {
 public:
-  static ErrorCode OpenFile(std::string const &path, uint32_t flags,
-                            uint32_t mode, int &fd);
-  static ErrorCode CloseFile(int fd);
+  ErrorCode OpenFile(std::string const &path, uint32_t flags, uint32_t mode,
```
sas: I like that you're making these non-`static`. Can you merge this work (of making it non-`static`) in the commit that introduces the class and make sure we have proper RTTI semantics on this?

Headers/DebugServer2/GDBRemote/DummySessionDelegateImpl.h
fjricci: This won't break with multiple open and close calls. A FileIO object isn't just one file (which is why I didn't call the class File).

Headers/DebugServer2/Host/Windows/FileIO.h
fjricci: By RTTI semantics, do you mean calling these `virtual`?

sas: WTF, I meant RAI, not RTTI :o

On Tue, May 31, 2016 at 3:19 PM, Francis Ricci <notifications@github.com>
wrote:

> In Headers/DebugServer2/Host/Windows/FileIO.h
> <https://github.com/facebook/ds2/pull/353#discussion_r65273462>:
>
> > @@ -25,13 +25,13 @@ namespace Windows {
> >
> >  class FileIO {
> >  public:
> > -  static ErrorCode OpenFile(std::string const &path, uint32_t flags,
> > -                            uint32_t mode, int &fd);
> > -  static ErrorCode CloseFile(int fd);
> > +  ErrorCode OpenFile(std::string const &path, uint32_t flags, uint32_t mode,
>
> By RTTI semantics, do you mean calling these virtual?
>
> —
> <https://github.com/facebook/ds2/pull/353/files/4e44056a3d3facb901289ae864a9e8bb629837dd..5e3cd77ac1ebcd965ff84fafe9f0b73b2b564021#r65273462>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADHbJU9qR0oVcsLqeD5dABx6jP_87_yks5qHLPhgaJpZM4ImqRQ>
> .
>



-- 
Stephane Sezer

Sources/GDBRemote/DummySessionDelegateImpl.cpp
```diff
@@ -11,9 +11,11 @@
 #include "DebugServer2/GDBRemote/DummySessionDelegateImpl.h"
 #include "DebugServer2/GDBRemote/Session.h"
 #include "DebugServer2/Host/Platform.h"
+#include "DebugServer2/Host/Platform.h"
 #include "DebugServer2/Utils/Log.h"
 
 using ds2::Host::Platform;
+using ds2::Host::Platform;
```
sas: Mhmmm?

fjricci: Just in case the first one didn't work.

Headers/DebugServer2/Host/File.h
```diff
@@ -0,0 +1,58 @@
...
+  ~File();
+
+public:
+  File(const File &other) = delete;
+  File &operator=(const File &other) = delete;
+
+public:
+  File(File &&other) { *this = std::move(other); }
+
+  File &operator=(File &&other) {
```
sas: Shouldn't you call your own destructor first?

fjricci: Didn't know I had to do that manually. But yeah, makes sense.

sas: Yeah otherwise you'll never `close()` the current FD you own.

On Mon, Jun 13, 2016 at 2:32 PM, Francis Ricci <notifications@github.com>
wrote:

> In Headers/DebugServer2/Host/File.h
> <https://github.com/facebook/ds2/pull/353#discussion_r66872451>:
>
> > +namespace ds2 {
> > +namespace Host {
> > +
> > +class File {
> > +public:
> > +  File(std::string const &path, uint32_t flags, uint32_t mode);
> > +  ~File();
> > +
> > +public:
> > +  File(const File &other) = delete;
> > +  File &operator=(const File &other) = delete;
> > +
> > +public:
> > +  File(File &&other) { *this = std::move(other); }
> > +
> > +  File &operator=(File &&other) {
>
> Didn't know I had to do that manually. But yeah, makes sense.
>
> —
> <https://github.com/facebook/ds2/pull/353/files/e78f7a117788527571efd48e401504161ff5090d..d75dab9690e2408866ce63d40f6d87e726e7de3e#r66872451>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADHbKpkZ9Z6AeTt9xusMqALR-bFXpicks5qLcx1gaJpZM4ImqRQ>
> .
>



-- 
Stephane Sezer

Sources/GDBRemote/DummySessionDelegateImpl.cpp
```diff
@@ -437,22 +437,26 @@ ErrorCode DummySessionDelegateImpl::onFileOpen(Session &,
...
-  //
-  // TODO(sas): Would be nice to have a table storing the FDs that were opened
-  // by the debugger to prevent Platform::CloseFile() getting called on
-  // arbitrary file descriptors.
-  //
-  if (!Platform::CloseFile(fd))
-    return Platform::TranslateError();
-  else
-    return kSuccess;
+  if (!_openFiles.count(fd)) {
```
sas: I'd use `_openFiles.find(fd)` here, and then use the resulting iterator in `_openFiles.erase()`. That way you only do one lookup in the container.

Sources/Host/POSIX/File.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace ds2 {
+namespace Host {
+
+File::File(std::string const &path, uint32_t flags, uint32_t mode) {
+  _fd = ::open(path.c_str(), flags, mode);
+  _lastError = (_fd < 0) ? Platform::TranslateError() : kSuccess;
+}
+
+File::~File() { ::close(_fd); }
```
sas: Doesn't matter much here, but it makes more sense to check first if `_fd >= 0` (or just call your own `valid()` method).

Sources/Host/Windows/File.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace ds2 {
+namespace Host {
+
+File::File(std::string const &path, uint32_t flags, uint32_t mode) {
+  _fd = -1;
+  _lastError = kErrorUnsupported;
+}
+
+File::~File() {}
```
sas: `File::~File() = default;`

Headers/DebugServer2/Host/File.h
```diff
@@ -45,6 +45,9 @@ class File {
   }
 
 public:
+  virtual ErrorCode pread(std::string &buf, uint64_t &count, uint64_t offset);
```
sas: If this is going to be a virtual method, then your destructor should be virtual as well.

Headers/DebugServer2/Host/File.h
fjricci: Actually, I don't think you can do that: http://stackoverflow.com/questions/32796739/can-i-call-destructor-in-move-assignment-operator

fjricci: I think you just have to do the close manually

Sources/Host/POSIX/File.cpp
```diff
@@ -23,5 +24,31 @@ File::File(std::string const &path, uint32_t flags, uint32_t mode) {
 }
 
 File::~File() { ::close(_fd); }
+
+ErrorCode File::pread(std::string &buf, uint64_t &count, uint64_t offset) {
+  if (_fd < 0) {
```
sas: You should use your own `valid()` method, and just return `kErrorInvalidHandle` if it's `false`.

Sources/Host/POSIX/File.cpp
```diff
@@ -23,5 +24,31 @@ File::File(std::string const &path, uint32_t flags, uint32_t mode) {
...
+    return _lastError = kErrorInvalidHandle;
+  }
+
+  if (offset > std::numeric_limits<off_t>::max()) {
+    return _lastError = kErrorInvalidArgument;
+  }
+
+  off_t offArg = static_cast<off_t>(offset);
+  size_t countArg = static_cast<size_t>(count);
```
sas: For lines 37 and 38, you should use `auto` instead of repeating the type both in the variable declaration and the `static_cast` parameter.

Sources/Host/POSIX/File.cpp
```diff
@@ -23,5 +24,31 @@ File::File(std::string const &path, uint32_t flags, uint32_t mode) {
 }
 
 File::~File() { ::close(_fd); }
+
+ErrorCode File::pread(std::string &buf, uint64_t &count, uint64_t offset) {
```
sas: Seems like using a `ByteVector` instead of `std::string` here makes more sense.

Sources/Host/POSIX/File.cpp
```diff
@@ -23,5 +24,31 @@ File::File(std::string const &path, uint32_t flags, uint32_t mode) {
...
+  if (offset > std::numeric_limits<off_t>::max()) {
+    return _lastError = kErrorInvalidArgument;
+  }
+
+  off_t offArg = static_cast<off_t>(offset);
+  size_t countArg = static_cast<size_t>(count);
+
+  buf.resize(countArg);
+
+  ssize_t nRead = ::pread(_fd, &buf[0], countArg, offArg);
```
sas: If you use a `ByteVector`, you can use `buf.data()` instead of `&buf[0]`.

Sources/GDBRemote/DummySessionDelegateImpl.cpp
```diff
@@ -459,9 +459,15 @@ ErrorCode DummySessionDelegateImpl::onFileClose(Session &session, int fd) {
   return kSuccess;
 }
 
-ErrorCode DummySessionDelegateImpl::onFileRead(Session &, int, size_t, uint64_t,
-                                               std::string &) {
-  return kErrorUnsupported;
+ErrorCode DummySessionDelegateImpl::onFileRead(Session &session, int fd,
+                                               uint64_t &count, uint64_t offset,
+                                               std::string &buffer) {
+  auto it = _openFiles.find(fd);
```
sas: This is a good way to do the lookup :p

Headers/DebugServer2/Utils/String.h
```diff
@@ -28,6 +29,13 @@ template <typename T> static inline std::string ToString(T val) {
   return os.str();
 }
 
+// For vector types, we need a special ToString implementation
+template <typename T> static inline std::string ToString(std::vector<T> vec) {
```
sas: `vec` should be a const reference here.

Headers/DebugServer2/Utils/HexValues.h
```diff
@@ -52,8 +52,7 @@ static inline std::string HexToString(std::string const &str) {
   std::string result;
   DS2ASSERT(str.size() % 2 == 0);
   for (size_t n = 0; n < str.size(); n += 2) {
-    result +=
-        static_cast<char>((HexToNibble(str[n]) << 4) | HexToNibble(str[n + 1]));
+    result += static_cast<char>(HexToByte(&str[n]));
```
sas: 👍

Sources/Host/POSIX/File.cpp
fjricci: Taken care of in a later commit.

Sources/Host/POSIX/File.cpp
fjricci: Gets changed in a couple of commits

Headers/DebugServer2/Host/File.h
```diff
@@ -0,0 +1,56 @@
...
+public:
+  File(std::string const &path, uint32_t flags, uint32_t mode);
+  ~File();
+
+public:
+  File(const File &other) = delete;
+  File &operator=(const File &other) = delete;
+
+public:
+  File(File &&other) { *this = std::move(other); }
```
sas: Initialize yourself before calling the move assignment operator.

Sources/Host/Windows/File.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "DebugServer2/Host/File.h"
+#include "DebugServer2/Host/Platform.h"
+#include "DebugServer2/Host/Windows/ExtraWrappers.h"
+
+namespace ds2 {
+namespace Host {
+
+File::File(std::string const &path, uint32_t flags, uint32_t mode) {
+  _fd = -1;
```
sas: Initializer list.

Headers/DebugServer2/GDBRemote/ProtocolHelpers.h
```diff
@@ -12,19 +12,72 @@
...
+  }
 
-std::string Uncompress(std::string const &data);
+  if (count == 0)
+    return data;
+
+  return ss.str();
+}
+
+template <typename T> std::string Compress(T const &data);
```
sas: `Compress` and `Uncompress` seem like they could be deleted.

Headers/DebugServer2/GDBRemote/ProtocolHelpers.h
```diff
@@ -12,19 +12,72 @@
...
+template <typename T> std::string Unescape(T const &data) {
+  std::ostringstream ss;
+  size_t first = 0;
+  size_t last = 0;
+  size_t count = 0;
+  size_t lastp = std::string::npos;
 
-std::string Compress(std::string const &data);
+  while ((last = data.find('}', last)) != std::string::npos) {
+    ss << data.substr(first, last - first);
```
sas: Won't work with `ByteVector`.

Sources/GDBRemote/SessionBase.cpp
```diff
@@ -16,6 +16,7 @@
 #include "DebugServer2/Utils/HexValues.h"
 #include "DebugServer2/Utils/Log.h"
 
+#include <algorithm>
```
sas: No.

Headers/DebugServer2/Utils/HexValues.h
```diff
@@ -39,11 +39,11 @@ static inline uint8_t HexToByte(char const *chars) {
   return (HexToNibble(chars[0]) << 4) | HexToNibble(chars[1]);
 }
 
-static inline std::string StringToHex(std::string const &str) {
-  std::string result;
-  for (char n : str) {
-    result += NibbleToHex(n >> 4);
-    result += NibbleToHex(n & 15);
+template <typename T> static inline T ToHex(T const &vec) {
```
sas: You should always return an `std::string`.

Sources/GDBRemote/Session.cpp
```diff
@@ -3114,9 +3115,10 @@ void Session::Handle_vFile(ProtocolInterpreter::Handler const &,
     }
 
     size_t nwritten;
+    uint8_t *bytePtr = reinterpret_cast<uint8_t *>(eptr);
```
sas: Use `auto` for the variable declaration.

Sources/GDBRemote/Session.cpp
```diff
@@ -3221,8 +3223,9 @@ void Session::Handle_vFlashWrite(ProtocolInterpreter::Handler const &,
     return;
   }
 
-  ErrorCode error =
-      _delegate->onFlashWrite(*this, address, std::string(eptr, length));
+  uint8_t *bytePtr = reinterpret_cast<uint8_t *>(eptr);
```
sas: Ditto.

Headers/DebugServer2/Host/File.h
```diff
@@ -0,0 +1,58 @@
...
+public:
+  File(std::string const &path, uint32_t flags, uint32_t mode);
+  ~File();
+
+public:
+  File(const File &other) = delete;
+  File &operator=(const File &other) = delete;
+
+public:
+  File(File &&other) : _fd(-1), _lastError(kErrorUnsupported) {
```
sas: Small nit: `kErrorUnknown` or `kErrorInvalidHandle` make more sense than `Unsupported`.

sas:
@facebook-github-bot shipit

fjricci:
Fixed nit and committed as 5acd904..056c504

PR #354, opened 2016-05-25
Respond to qSymbol query packets correctly

fjricci:


sas:
I think the problem with this is that debuggers (lldb and probably gdb) send a query packet with an empty symbol to check whether we support this packet or not. This commit will make us advertise support, but we don't actually support anything after that.

fjricci:
‘qSymbol::’
Notify the target that gdb is prepared to serve symbol lookup requests. Accept requests from the target for the values of symbols.
Reply:

‘OK’
The target does not need to look up any (more) symbols.

fjricci:
From looking at the remote packet logs, this doesn't appear to affect the behavior of gdb or lldb (since it means "I don't need you to do anything further"). But it also seems more correct to me (since we haven't implemented the packet precisely because we never need it to do anything).

sas:
Scratch everything I said. This commit is good to go.

sas:
@facebook-github-bot shipit

PR #355, opened 2016-05-26
Handle uncaught WinRT exception

a20012251:


sas:
@facebook-github-bot shipit

PR #357, opened 2016-05-27
Force BP size to 2 on Windows ARM.

sas:
Windows ARM (more precisely, WinPhone ARM) is a pure thumb environment,
so if the debugger is requesting a breakpoint of size 4, it's probably
an older/buggy version and we should enforce size 2.

### DIFF
--- a/Sources/Architecture/ARM/SoftwareBreakpointManager.cpp
+++ b/Sources/Architecture/ARM/SoftwareBreakpointManager.cpp
@@ -113,6 +113,16 @@ bool SoftwareBreakpointManager::hit(Target::Thread *thread) {
 
 void SoftwareBreakpointManager::getOpcode(uint32_t type,
                                           std::string &opcode) const {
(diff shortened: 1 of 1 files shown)


strager:
Aren't 4-byte breakpoints valid for THUMB-2?

sas:
They are, but they are identified with `type 3` in the GDB protocol. `type
4` always means standard 4-bytes ARM breakpoint.

On Fri, May 27, 2016 at 3:14 PM, strager <notifications@github.com> wrote:

> Aren't 4-byte breakpoints valid for THUMB-2?
>
> —
> <https://github.com/facebook/ds2/pull/357#issuecomment-222264601>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AADHbMEdmrJs-ttptvazEpOE0sG2sKfsks5qF2yogaJpZM4Io5rD>
> .
>



-- 
Stephane Sezer

a20012251:
@facebook-github-bot shipit

PR #361, opened 2016-06-06
Make sure to read CPUState on the correct thread on memory alloc/dealloc

fjricci:
Fixes thumb/arm mismatch if different threads are in different modes.


strager:
+2

I hate Linux threads.

sas:
@facebook-github-bot shipit

PR #362, opened 2016-06-07
Update Darwin gcc version to match homebrew

fjricci:


fjricci:
@kokarez I think this is required now, unless there's a way to specify an older version when you do a brew install

fjricci:
Committed one commit as 6a9e5bb

PR #364, opened 2016-06-07
Silence -Wunused-but-set-parameter on gcc 6+.

sas:


fjricci:
@facebook-github-bot shipit

PR #365, opened 2016-06-07
Implement a custom DebugBreakProcess

a20012251:


Headers/DebugServer2/Host/Windows/ExtraWrappers.h
```diff
@@ -375,10 +375,6 @@ WINBASEAPI BOOL WINAPI VirtualFreeEx(
   _In_ DWORD  dwFreeType
 );
 
-WINBASEAPI BOOL WINAPI DebugBreakProcess(
```
sas: Yay!

Sources/Target/Windows/Process.cpp
```diff
@@ -150,14 +157,18 @@ BOOL Process::debugBreakProcess() {
...
+                         &threadId) == NULL) {
     return FALSE;
   }
   return TRUE;
-#else
-  return DebugBreakProcess(_handle);
-#endif
 }
 
 ErrorCode Process::interrupt() {
```
sas: This function does nothing besides call `Process::debugBreakProcess`. Can you move the contents of `Process::debugBreakProcess` into `Process::interrupt` and get rid of the former?

When you do that, you'll probably have to make each error checking block return `Platform::TranslateError()` instead of `TRUE` or `FALSE`.

Sources/Target/Windows/Process.cpp
```diff
@@ -105,15 +122,7 @@ ErrorCode Process::detach() {
...
 }
 
-#if defined(ARCH_X86)
-static uint8_t const debugBreakCode[] = {
-    0xCC,                         // 00: int 3
-    0xB8, 0x00, 0x00, 0x00, 0x00, // 01: mov eax, RtlExitUserThread address
-    0xFF, 0xD0                    // 06: call eax
-};
-
 BOOL Process::writeDebugBreakCode(uint64_t address) {
```
sas: I like how this function became 100% architecture agnostic, and the `#ifdef` is limited to `debugBreakCode`.

Sources/Target/Windows/Process.cpp
```diff
@@ -27,6 +27,23 @@ using ds2::Utils::Stringify;
 
 #define super ds2::Target::ProcessBase
 
+#if defined(ARCH_ARM)
+static uint8_t const debugBreakCode[] = {
```
sas: Nit: I think global variables are marked with a `g` prefix in this code. e.g.: `gDebugBreakCode`.

Sources/Target/Windows/Process.cpp
```diff
@@ -104,10 +122,52 @@ ErrorCode Process::detach() {
   return kSuccess;
 }
 
+BOOL Process::writeDebugBreakCode(uint64_t address) {
+  FARPROC exitThreadAddress =
+      GetProcAddress(GetModuleHandleA("ntdll"), "RtlExitUserThread");
```
sas: Is it correct that `GetModuleHandleA` cannot fail here? I think you have to have that assumption correct otherwise this code may fail. (Either in ds2, or with a segfault in the inferior if `GetProcAddress` returns `0`).

Sources/Target/Windows/Process.cpp
```diff
@@ -104,10 +122,52 @@ ErrorCode Process::detach() {
...
+  if (exitThreadAddress == NULL) {
+    return FALSE;
+  }
+
+  ByteVector codestr(&debugBreakCode[0],
+                     &debugBreakCode[sizeof(debugBreakCode)]);
+  *reinterpret_cast<uint32_t *>(&codestr[kRtlExitUserThreadOffset]) =
```
sas: This code assumes the endianness of ds2 and the endianness of the inferior are the same. I think this is a safe assumption.

Sources/Target/Windows/Process.cpp
```diff
@@ -104,10 +122,52 @@ ErrorCode Process::detach() {
   return kSuccess;
 }
 
+BOOL Process::writeDebugBreakCode(uint64_t address) {
```
sas: It may or may not be useful to make `Process::writeDebugBreakCode` return an `ErrorCode` instead. That way you have slightly more info on what failed after calling `writeMemory` for instance.

Sources/Target/Windows/Process.cpp
```diff
@@ -104,10 +122,52 @@ ErrorCode Process::detach() {
...
+}
+
 ErrorCode Process::interrupt() {
-  BOOL result = DebugBreakProcess(_handle);
-  if (!result)
+  SYSTEM_INFO info;
+  GetSystemInfo(&info);
+
+  uint64_t address;
+  if (allocateMemory(info.dwPageSize, kProtectionExecute | kProtectionWrite,
```
sas: I think `Process::allocateMemory` returns an `ErrorCode` already, no need to drop it and then call `Platform::TranslateError()` again. (I'm not even 100% sure you can call `GetLastError()` multiple times in a row and get the expected result).

Sources/Target/Windows/Process.cpp
```diff
@@ -104,10 +122,52 @@ ErrorCode Process::detach() {
...
+                     &address) != kSuccess) {
+    return Platform::TranslateError();
+  }
+
+  if (!writeDebugBreakCode(address)) {
     return Platform::TranslateError();
+  }
+
+  DWORD threadId;
+#if defined(ARCH_ARM)
```
sas: Nit: you could do 
```
#if defined(ARCH_ARM)
    // comment...
    address |= 1ULL;
#endif
```

Shorter and reduces the number of unused temporary variables.

Sources/Target/Windows/Process.cpp
a20012251: if any of these functions fail exitThreadAddress will be NULL anyway

PR #366, opened 2016-06-09
Clear thumb bit from Windows register context.

sas:
This confuses the debugger, and is not what the CPU's PC register
actually contains. Get rid of it so everything works in sync properly.


a20012251:
@facebook-github-bot shipit

PR #367, opened 2016-06-10
Fix breakpoints on ARM

a20012251:


Sources/Architecture/ARM/SoftwareBreakpointManager.cpp
```diff
@@ -109,7 +109,22 @@ void SoftwareBreakpointManager::enumerate(
 bool SoftwareBreakpointManager::hit(Target::Thread *thread) {
   ds2::Architecture::CPUState state;
   thread->readCPUState(state);
-  return super::hit(state.pc());
+  state.setPC(state.pc() - 2);
```
sas: I think this whole block should happen only `#if defined(OS_WIN32)`. Linux seems to handle the PC adjustment by itself.
If it's not too annoying, can you test that behavior yourself?

Sources/Architecture/ARM/SoftwareBreakpointManager.cpp
```diff
@@ -109,7 +109,22 @@ void SoftwareBreakpointManager::enumerate(
...
+    //
+    // Move the PC back to the instruction
+    //
+    if (thread->writeCPUState(state) != kSuccess) {
+      abort();
+    }
+
+    uint64_t ex = state.pc();
+    thread->readCPUState(state);
+    DS2ASSERT(ex == state.pc());
```
sas: I don't think we should call `readCPUState()` again just to check this. As @fjricci pointed out, this check/assertion is pretty useless.

Sources/BreakpointManager.cpp
```diff
@@ -162,11 +162,6 @@ bool BreakpointManager::hit(Address const &address) {
 
   it->second.type =
       static_cast<Type>(it->second.type & ~kTypeTemporaryUntilHit);
-  if (!it->second.type) {
```
sas: Removing this is good. Can you just add a small comment before line 163 to mention that the breakpoint will be `erase`'d when we call `BreakpointManager::disable()` if its `type` field goes to 0?

sas:
I think this PR is good overall. You are introducing a small bug for non-Windows ARM debugging, but it should be easy to fix before checking this in.

sas:
@facebook-github-bot shipit

Sources/Architecture/ARM/SoftwareBreakpointManager.cpp
```diff
@@ -109,7 +109,21 @@ void SoftwareBreakpointManager::enumerate(
...
   ds2::Architecture::CPUState state;
   thread->readCPUState(state);
+#if defined(OS_WIN32)
+  state.setPC(state.pc() - 2);
+  if (super::hit(state.pc())) {
+    //
+    // Move the PC back to the instruction
+    //
+    if (thread->writeCPUState(state) != kSuccess) {
+      abort();
```
fjricci: Is abort() better than DS2ASSERT?

PR #368, opened 2016-06-13
Enable lldb tests which are XFAIL with lldb-server but pass with ds2

fjricci:


sas:
@facebook-github-bot shipit

PR #369, opened 2016-06-13
Fix incorrect ARM CPUState register definition.

sas:
ip and sp were swapped. This worked fine on Linux because we didn't use
the named registers directly.


a20012251:
@facebook-github-bot shipit

PR #370, opened 2016-06-14
Don't use unnecessary unique ptr in platform mode

fjricci:
The destructor for SessionBase will always free its channel in
the destructor, prevent double-frees


sas:
@facebook-github-bot shipit

fjricci:
Used the merge and squash button! 😆

PR #372, opened 2016-06-15
Add platform mode testing to ds2

fjricci:


Support/Scripts/run-lldb-tests.sh
```diff
@@ -125,6 +125,12 @@ else
   fi
 fi
 
+if [[ "${PLATFORM-}" = "1" ]]; then
```
sas: I think you should move this block to right before the final `exec` call, and use the strace stuff we have on lines 133-142. Probably do something like `ds2_path=...` (with the shell script or with just the binary), and then if you have `${PLATFORM-}`, just run $ds2_path "p" &, else `export LLDB_DEBUGSERVER_PATH="$ds2_path"`.

Support/Scripts/run-lldb-tests.sh
```diff
@@ -125,20 +125,34 @@ else
...
+
+if [[ "${PLATFORM-}" = "1" ]]; then
+  args="$args --platform-name=remote-linux --platform-url=connect://localhost:12345 --no-multiprocess"
+  ds2_args="p"
+
+  if $opt_log; then
+    ds2_args="$ds2_args --remote-debug --log-file=$build_dir/ds2.log"
+  fi
+
+  $ds2_path "$ds2_args" &
```
sas: This is why your tests are failing I think. You should remove the quotes here.

sas:
There's a quoting bug in your script. Other than that, this is good to go.

@facebook-github-bot shipit

Support/Scripts/run-lldb-tests.sh
```diff
@@ -18,7 +18,14 @@ cherry_pick_patches() {
...
   local testingPath="$top/Support/Testing"
-  for p in "$testingPath"/Patches/lldb/*.patch; do
+  for p in "$testingPath"/Patches/lldb-disable/*.patch; do
+    echo "Applying $p"
+    patch -d "$lldb_path" -p1 <"$p"
+  done
+
+  # Enabled tests
+  local testingPath="$top/Support/Testing"
+  for p in "$testingPath"/Patches/lldb-enable/*.patch; do
```
sas: `for p in "$testingPath"/Patches/lldb-{enable,disable}/*.patch; do`

sas:
Looks good to me except for the useless loop duplication.

PR #373, opened 2016-06-15
Skip some function declarations when running doxygen.

sas:
This skips all the extra wrappers we have for Windows and Linux. These
are just redefinitions of system headers so we don't need to include
them in our documentation.


fjricci:
@facebook-github-bot shipit

PR #375, opened 2016-06-15
Implement qPlatform_mkdir

fjricci:


Headers/DebugServer2/GDBRemote/Session.h
```diff
@@ -143,8 +143,8 @@ class Session : public SessionBase {
   void Handle_qOffsets(ProtocolInterpreter::Handler const &,
                        std::string const &);
   void Handle_qP(ProtocolInterpreter::Handler const &, std::string const &);
-  void Handle_qPlatform_IO_MkDir(ProtocolInterpreter::Handler const &,
-                                 std::string const &);
+  void Handle_qPlatform_mkdir(ProtocolInterpreter::Handler const &,
```
sas: Can you make sure this is not a case of gdb vs. lldb? I find it surprising that we were handling a packet named differently, with a different format, etc.

Sources/Host/POSIX/File.cpp
```diff
@@ -54,5 +55,28 @@ ErrorCode File::pread(ByteVector &buf, uint64_t &count, uint64_t offset) {
 
   return _lastError = kSuccess;
 }
+
+ErrorCode File::createDirectory(std::string const &path, uint32_t flags) {
+  size_t pos = 0;
+
+  // Each parent directory must be created individually, if it does not exist
+  while ((pos = path.find('/', pos + 1)) != std::string::npos) {
```
sas: Nit: use a `for` loop and avoid the external variable declaration. (also, the initialization of that variable is not required).

Sources/Host/POSIX/File.cpp
```diff
@@ -54,5 +55,28 @@ ErrorCode File::pread(ByteVector &buf, uint64_t &count, uint64_t offset) {
...
+    std::string partialPath = path.substr(0, pos);
+
+    if (!Platform::IsFilePresent(partialPath.c_str())) {
```
sas: You shouldn't do `IsFilePresent`. Just `::mkdir()` the path, and if it fails, check if `errno` is `EEXIST`.

Sources/Host/POSIX/File.cpp
```diff
@@ -54,5 +55,28 @@ ErrorCode File::pread(ByteVector &buf, uint64_t &count, uint64_t offset) {
...
+      if (::mkdir(partialPath.c_str(), flags) < 0) {
+        return Platform::TranslateError();
+      }
+    }
+  }
+
+  if (!Platform::IsFilePresent(path.c_str())) {
```
sas: I think if you don't do a `for` loop but keep a `while` loop, you can do a `do {} while();` loop to avoid doing this code twice.

Sources/Host/Windows/File.cpp
```diff
@@ -23,5 +23,10 @@ File::~File() = default;
 ErrorCode File::pread(ByteVector &buf, uint64_t &count, uint64_t offset) {
   return kErrorUnsupported;
 }
+
+static ErrorCode File::createDirectory(std::string const &path,
+                                       uint32_t flags) {
+  return kErrorUnsupported;
```
sas: Still solid 😆

Headers/DebugServer2/GDBRemote/Session.h
fjricci: It's not. I did a google search for the former, and the only results were some old lldb patches. A grep in the gdb and lldb sources for the former return nothing, and only the `qPlatform_mkdir` shows up in their docs. My assumption is that they started implementing and then decided it was a bad name.

Sources/Host/POSIX/File.cpp
fjricci: Why is the initialization not required? You have to tell find to start at the beginning of the string.

Headers/DebugServer2/GDBRemote/Session.h
sas: That makes sense. Thanks for checking that stuff.

On Wed, Jun 15, 2016 at 4:01 PM, Francis Ricci <notifications@github.com>
wrote:

> In Headers/DebugServer2/GDBRemote/Session.h
> <https://github.com/facebook/ds2/pull/375#discussion_r67261613>:
>
> > @@ -143,8 +143,8 @@ class Session : public SessionBase {
> >    void Handle_qOffsets(ProtocolInterpreter::Handler const &,
> >                         std::string const &);
> >    void Handle_qP(ProtocolInterpreter::Handler const &, std::string const &);
> > -  void Handle_qPlatform_IO_MkDir(ProtocolInterpreter::Handler const &,
> > -                                 std::string const &);
> > +  void Handle_qPlatform_mkdir(ProtocolInterpreter::Handler const &,
>
> It's not. I did a google search for the former, and the only results were
> some old lldb patches. A grep in the gdb and lldb sources for the former
> return nothing, and only the qPlatform_mkdir shows up in their docs. My
> assumption is that they started implementing and then decided it was a bad
> name.
>
> —
> You are receiving this because you were assigned.
> <https://github.com/facebook/ds2/pull/375/files/687555e04395f15d83108877ce62d7c6805a3aa9#r67261613>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADHbHDbpDTj3KSA3CkV3lAA7nCj6zVxks5qMIRUgaJpZM4I229L>
> .
>



-- 
Stephane Sezer

Sources/Host/POSIX/File.cpp
sas: Yeah the initialization not being required is me not being able to read
code, ignore that comment.

On Wed, Jun 15, 2016 at 4:02 PM, Francis Ricci <notifications@github.com>
wrote:

> In Sources/Host/POSIX/File.cpp
> <https://github.com/facebook/ds2/pull/375#discussion_r67261680>:
>
> > @@ -54,5 +55,28 @@ ErrorCode File::pread(ByteVector &buf, uint64_t &count, uint64_t offset) {
> >
> >    return _lastError = kSuccess;
> >  }
> > +
> > +ErrorCode File::createDirectory(std::string const &path, uint32_t flags) {
> > +  size_t pos = 0;
> > +
> > +  // Each parent directory must be created individually, if it does not exist
> > +  while ((pos = path.find('/', pos + 1)) != std::string::npos) {
>
> Why is the initialization not required? You have to tell find to start at
> the beginning of the string.
>
> —
> You are receiving this because you were assigned.
> <https://github.com/facebook/ds2/pull/375/files/687555e04395f15d83108877ce62d7c6805a3aa9..29a134830a0744f8aad6e24930060c19ebd6684f#r67261680>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADHbMsA9AkWwVpwxetelkTrhdXjK6Lkks5qMISEgaJpZM4I229L>
> .
>



-- 
Stephane Sezer

sas:
@facebook-github-bot shipit

fjricci:
Committed as e8fbd70..b552c57

PR #376, opened 2016-06-16
Implement vFile:open flag parsing

fjricci:


Headers/DebugServer2/Constants.h
```diff
@@ -76,6 +76,22 @@ enum /*Protection*/ {
   kProtectionRead = (1 << 2)
 };
 
+//
+// Open flags
+//
+
+enum OpenMode {
+  kModeRead = 0,
```
sas: This should be `kModeRead = (1 << 0),`

Sources/Host/POSIX/File.cpp
```diff
@@ -18,8 +18,36 @@
...
+    flags |= O_APPEND;
+  if (ds2Flags & kModeTruncate)
+    flags |= O_TRUNC;
+  if (ds2Flags & kModeNonBlocking)
+    flags |= O_NONBLOCK;
+  if (ds2Flags & kModeCreate)
+    flags |= O_CREAT;
+  if (ds2Flags & kModeNewOnly)
+    flags |= O_EXCL;
+#if defined(OS_LINUX)
```
sas: Here you should probably do `#if defined(O_NOFOLLOW)` instead, because I suppose systems other than Linux support that flag.

Also, adding an `#else` block and returning an error like unsupported or invalidargument *might* make sense?

Sources/Host/POSIX/File.cpp
```diff
@@ -18,8 +18,36 @@
 namespace ds2 {
 namespace Host {
 
+static uint32_t convertFlags(uint32_t ds2Flags) {
```
sas: Instead of uint32_t I think you can use the enum type name.

fjricci: No longer possible, since I added an error check (which returns 0)

sas:
Nits: You can make the convert function in `Session.cpp` be non-static, and you should use `protocolFlags` instead of `oldFlags`. Other than that, this is good.

@facebook-github-bot shipit

fjricci:
Committed as f81d090..2a49eb2

PR #377, opened 2016-06-16
Fix escape character when unescaping packets

fjricci:


a20012251:
@facebook-github-bot shipit

fjricci:
Retesting because appveyor had an infra issue.

PR #381, opened 2016-06-16
Use OS_POSIX instead of listing all POSIX systems.

sas:


a20012251:
@facebook-github-bot shipit

PR #386, opened 2016-06-17
Don't hex encode vFile count responses

fjricci:


sas:
@facebook-github-bot shipit

How did this work before?

fjricci:
Well, this packet wasn't being used at all until I implemented the underlying vFile stuff. I was just dumb and assumed that the response part was correct 😆

fjricci:
(As in, the response stuff was existing code that I didn't verify until now)

fjricci:
Turns out that lldb uses decimal but gdb uses hex. Because why not.

a20012251:
@facebook-github-bot shipit

fjricci:
Fixed nit on variable name and committed as a447204262690658380b90f1782383337d4aa4f9

PR #390, opened 2016-06-30
Fix LSAN crash in platform mode ds2

fjricci:


sas:
@facebook-github-bot shipit

fjricci:
Committed as 0eaf2de..3dbc2e4

PR #396, opened 2016-07-05
Get rid of `[[noreturn]]` on `onExitServer`.

sas:
Not all session delegates implement `onExitServer` and some might return
an error, using `noreturn` means not all overrides have the same
interface.


fjricci:
@facebook-github-bot shipit

sas:
Committed as c30db2a44c905ac91b24e057dae2652935b738a2.

PR #398, opened 2016-07-05
Update README for Windows builds

fjricci:


README.md
```diff
@@ -102,10 +102,13 @@ Then, as with linux, use CMake to generate the build system, then use msbuild
 ```sh
 cd ds2
 mkdir build && cd build
-cmake ..
+cmake -G "Visual Studio 14 2015" -DCMAKE_TOOLCHAIN_FILE=../Support/CMake/Toolchain-WinStore.cmake ..
```
sas: You should use `Toolchain-Windows.cmake` instead of `Toolchain-WinStore.cmake` here. Most users will just want a standard `ds2.exe` to debug stuff.

If you want to add a section for WinStore/WinPhone, feel free to do it, that would be very useful :)

fjricci: Sounds good.

PR #405, opened 2016-07-27
Implement packet qFileLoadAddress for windows

a20012251:


Sources/GDBRemote/DebugSessionImpl.cpp
```diff
@@ -458,6 +458,19 @@ ErrorCode DebugSessionImplBase::onQuerySharedLibrariesInfoAddress(
 #endif
 }
 
+ErrorCode DebugSessionImplBase::onQueryFileLoadAddress(
+    Session &session, std::string const &file_path, Address &address) {
+  if (_process == nullptr) {
+    return kErrorProcessNotFound;
+  }
+
+#if defined(OS_WIN32)
```
sas: Instead of doing this, you should put `getFileLoadAddress` in `Process.h` and have it return `kErrorUnsupported` on non-windows platforms. That way, this function can stay platform-agnostic, and we can avoid introducing differences in the interfaces of each OS's `Process` class.

Sources/GDBRemote/Session.cpp
```diff
@@ -1541,6 +1542,32 @@ void Session::Handle_qCRC(ProtocolInterpreter::Handler const &,
...
+void Session::Handle_qFileLoadAddress(ProtocolInterpreter::Handler const &,
+                                      std::string const &args) {
+  if (args.empty()) {
+    sendError(kErrorInvalidArgument);
+    return;
+  }
+
+  Address address;
+  ErrorCode error = _delegate->onQueryFileLoadAddress(
+      *this, std::string("COMProcessRunner.exe"), address);
```
sas: Mhm? Shouldn't you use `args` here?

Sources/Target/Windows/Process.cpp
```diff
@@ -350,6 +351,22 @@ ErrorCode Process::writeMemory(Address const &address, void const *data,
   return kSuccess;
 }
 
+ErrorCode Process::getFileLoadAddress(std::string const &file_path,
```
sas: Seems like this function doesn't do anything windows-specific. Maybe you can just pull it to `ProcessBase` as is?

Headers/DebugServer2/Target/Windows/Process.h
```diff
@@ -59,6 +59,7 @@ class Process : public ds2::Target::ProcessBase,
   }
 
 public:
+  ErrorCode getFileLoadAddress(std::string const &file_path, Address &address);
```
sas: You might want to make this have a similar interface to `enumerateSharedLibraries`, just for consistency.

sas:
Cool stuff. Just a couple of things to change for the interface, but otherwise this is good :)

Sources/GDBRemote/Session.cpp
a20012251: ahahah, i'm so stupid

Headers/DebugServer2/Target/Windows/Process.h
a20012251: I don't get this

sas: Instead of having `getFileLoadAddress` you can have something like `enumerateMappedFiles` or `enumerateLoadedFiles`, that would take a callback just like `enumerateSharedLibraries`. Then you can call `enumerateLoadedFiles` from `DebugSessionImpl` and put whatever logic for finding mapped files there (e.g.: taking only the basename, etc).

Headers/DebugServer2/Types.h
```diff
@@ -332,6 +332,11 @@ struct SharedLibraryInfo {
   } svr4;
   std::vector<uint64_t> sections;
 };
+
+struct MappedFileInfo {
+  std::string path;
+  uint64_t baseAddress;
```
sas: This should also have a size field of some sort.

Headers/DebugServer2/Target/Darwin/MachOProcess.h
```diff
@@ -31,7 +31,7 @@ class MachOProcess : public ds2::Target::POSIX::Process {
 public:
   virtual ErrorCode getSharedLibraryInfoAddress(Address &address);
   virtual ErrorCode enumerateSharedLibraries(
-      std::function<void(SharedLibraryInfo const &)> const &cb);
```
sas: If you add `override`, you should remove the `virtual` keyword here. You can skip this and we can do it in a different commit if you want.

Headers/DebugServer2/Target/POSIX/ELFProcess.h
```diff
@@ -30,7 +30,7 @@ class ELFProcess : public Process {
 public:
   virtual ErrorCode getSharedLibraryInfoAddress(Address &address);
   virtual ErrorCode enumerateSharedLibraries(
-      std::function<void(SharedLibraryInfo const &)> const &cb);
```
sas: Ditto.

Headers/DebugServer2/Target/Windows/Process.h
```diff
@@ -75,7 +75,7 @@ class Process : public ds2::Target::ProcessBase,
 
 public:
   ErrorCode enumerateSharedLibraries(
-      std::function<void(SharedLibraryInfo const &)> const &cb);
+      std::function<void(SharedLibraryInfo const &)> const &cb) override;
```
sas: This is good, without the `virtual`. Funny that this was never a virtual function in the first place.

Sources/GDBRemote/DebugSessionImpl.cpp
```diff
@@ -458,6 +458,28 @@ ErrorCode DebugSessionImplBase::onQuerySharedLibrariesInfoAddress(
...
+  }
+
+  ErrorCode error =
+      _process->enumerateMappedFiles([&](MappedFileInfo const &file) {
+        if (file.path == file_path ||
+            ds2::Utils::Basename(file.path) == file_path) {
+          address = Address(file.baseAddress);
+        }
+      });
+  if (error != kSuccess) {
```
sas: You can use one of my fancy macros for this I think? 😆

Sources/GDBRemote/Session.cpp
```diff
@@ -1541,6 +1542,32 @@ void Session::Handle_qCRC(ProtocolInterpreter::Handler const &,
...
+                                      std::string const &args) {
+  if (args.empty()) {
+    sendError(kErrorInvalidArgument);
+    return;
+  }
+
+  Address address;
+  ErrorCode error =
+      _delegate->onQueryFileLoadAddress(*this, HexToString(args), address);
+  if (error == kSuccess && !address.valid()) {
```
sas: Is this useful? Is it possible to have `onQueryFileLoadAddress` return `kSuccess` and for the `address` to be invalid?

Headers/DebugServer2/Target/ProcessBase.h
```diff
@@ -98,6 +98,12 @@ class ProcessBase {
                                size_t length, size_t *nread = nullptr) = 0;
   virtual ErrorCode writeMemory(Address const &address, void const *buffer,
                                 size_t length, size_t *nwritten = nullptr) = 0;
+  virtual ErrorCode enumerateSharedLibraries(
+      std::function<void(SharedLibraryInfo const &)> const &cb) = 0;
+
+public:
+  ErrorCode
```
sas: This should be `virtual` so that child classes can override it.

Sources/Target/Windows/Process.cpp
```diff
@@ -488,6 +488,9 @@ ErrorCode Process::enumerateSharedLibraries(
     // module itself.
     sl.sections.push_back(reinterpret_cast<uint64_t>(m));
 
+    // TODO: import GetModuleInformation to fetch the actual size
+    sl.sizeOfImage = -1;
```
sas: This should also be in the POSIX version, not only the Windows one.

Sources/Target/Common/ProcessBase.cpp
```diff
@@ -198,7 +198,7 @@ ErrorCode ProcessBase::enumerateMappedFiles(
   // TODO: this is only looking for modules loaded in memory, not all the
   // possible files
   return enumerateSharedLibraries([&](SharedLibraryInfo const &library) {
-    cb({library.path, library.sections[0]});
+    cb({library.path, library.sections[0], library.sizeOfImage});
```
sas: Also, sizeOfImage won't work for non-Windows platforms I think because the mappings are non-contiguous. You should probably remove `sizeOfImage` from the `SharedLibraryInfo` structure (but keep it in `MappedFileInfo`), and just pass `-1` here directly to the constructor of `MappedFileInfo`).

Headers/DebugServer2/Target/ProcessBase.h
```diff
@@ -102,7 +102,7 @@ class ProcessBase {
       std::function<void(SharedLibraryInfo const &)> const &cb) = 0;
 
 public:
-  ErrorCode
+  virtual ErrorCode
```
sas: Nit: this should be in the previous commit.

PR #406, opened 2016-08-02
Commit pages before writing to memory on windows

a20012251:


Sources/Target/Windows/Process.cpp
```diff
@@ -331,6 +331,18 @@ ErrorCode Process::readMemory(Address const &address, void *data, size_t length,
...
+  LPVOID allocError = VirtualAllocEx(
+      _handle,
+      reinterpret_cast<LPVOID>(address.value()),
+      length,
+      MEM_COMMIT,
+      PAGE_EXECUTE_READWRITE);
+  if (allocError == NULL) {
+    return Platform::TranslateError();
+  }
+
```
strager: This changes the "behaviour" of the memory (from the debuggee' perspective). Aside from turning an exception into a valid read/write (__chk_stack?), you're changing the protection of the memory.

This also will allocate memory if it was never allocated, which I'd imagine would hide bugs.

In what cases are we trying to write to virtual addresses without committed memory behind them?

a20012251: In fact, i'm changing the protection of memory, I will reuse the same protection it had before.
According to msdn, commiting an unallocated page doesn't allocate memory, the operation fails instead
I've seen that lldb sometimes tries to write some memory in a page and it fails unless the page is commited, I'll investigate more this issue

strager: > According to msdn, commiting an unallocated page doesn't allocate memory, the operation fails instead

Ah, I missed that. Confusing.

PR #410, opened 2016-08-03
Fix library size query for enumerateMappedFiles on POSIX.

sas:


a20012251:
hon hon hon

@facebook-github-bot shipit

PR #411, opened 2016-08-03
Make pages writable before actually writing on Windows

a20012251:


Headers/DebugServer2/Host/Windows/ExtraWrappers.h
```diff
@@ -31,6 +31,13 @@ extern "C" {
 
 #define UNLEN 256
 
+WINBASEAPI SIZE_T WINAPI VirtualQueryEx(
```
sas: Not a functional change, but can you put this next to `VirtualAllocEx` and friends?

Sources/Target/Windows/Process.cpp
```diff
@@ -44,6 +44,12 @@ static uint8_t const gDebugBreakCode[] = {
 static const int gRtlExitUserThreadOffset = 0x02;
 #endif
 
+static DWORD getPageSize() {
```
sas: I think if you rebase on top of master you can get rid of this and just use `Platform::GetPageSize()`.

Headers/DebugServer2/Target/Windows/Process.h
```diff
@@ -52,6 +52,16 @@ class Process : public ds2::Target::ProcessBase,
...
                         size_t *nwritten = nullptr) override;
 
+private:
+  ErrorCode makeMemoryWritable(Address const &address, size_t length,
+                               std::vector<MemoryRegionInfo> &modifiedRegions);
+  ErrorCode
+  restoreRegionsProtection(const std::vector<MemoryRegionInfo> &regions);
+  ErrorCode getMemoryRegionInfoInternal(Address const &address,
+                                        MemoryRegionInfo &info);
+  DWORD convertMemoryProtectionFromLLDB(uint32_t protection);
```
sas: These are not converting to/from LLDB memory protection types, they're converting from ds2 to windows type and vice versa, so you should just call them to/from Windows I think. There are other functions like these for file mode conversions in one of @fjricci's patches. Look at how he did that maybe?

a20012251: the naming makes sense, the other conversions were done inside a bigger function

a20012251:
the travis failure seems to be unrelated

fjricci:
Yeah, I broke stuff, you should rebase :)

sas:
Committed as d9ffba46547850f4d806b59074ac3ab2a2d2f4e7.

PR #412, opened 2016-08-03
Mark language as generic in travis config.

sas:
Without this, travis defaults to ruby for some reason, and does some
useless work at the beginning to support ruby testing.


a20012251:
lol
@facebook-github-bot shipit

PR #426, opened 2016-08-16
Add a declaration for VirtualQueryEx

sas:
We will need that and this is not available to Windows Store apps.


a20012251:
@facebook-github-bot shipit

PR #431, opened 2016-08-26
Detach from debugged process when closing connection

a20012251:
Specially on winphone, if we don't detach in this case, the debugged process keeps alive in a very weird state


sas:
@facebook-github-bot shipit

PR #433, opened 2016-08-29
Improve `CHK` and use it a bit more

sas:


a20012251:
@facebook-github-bot shipit

PR #437, opened 2016-09-22
Make it possible to install/run android emulators on macOS

sas:


Support/Scripts/install-android-emulator.sh
```diff
@@ -9,17 +9,29 @@
...
 
 if [ ! -f "$android_script" ] ; then
-  android_sdk_name="android-sdk_r24.4.1-linux"
-  wget "https://dl.google.com/android/$android_sdk_name.tgz" -P "/tmp/"
-  tar -zxf "/tmp/$android_sdk_name.tgz" -C "/tmp/"
+  android_sdk_package="android-sdk_${android_sdk_version}-${platform_name}.${package_extension}"
+  wget "https://dl.google.com/android/${android_sdk_package}" -P /tmp
+  case "${package_extension}" in
+    "zip") unzip -d /tmp "/tmp/${android_sdk_package}";;
+    *)     tar -C /tmp -xvf "/tmp/${android_sdk_package}";;
```
fjricci: I'd make this an explicit "tgz", in case we add other platforms, etc in the future

PR #441, opened 2016-09-30
Use `brew reinstall`

sas:
This prevents having to struggle with `brew install` then `brew update`,
and a bunch of `|| true` hacks.


fjricci:
Lgtm

PR #447, opened 2016-10-12
Fix coveralls coverage

fjricci:


fjricci:
Because I'm incompetent.

PR #448, opened 2016-10-13
Fix coverage on clang builds

fjricci:


CMakeLists.txt
```diff
@@ -58,7 +71,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang"
   endif ()
 
   if (COVERAGE)
-    set(COMMON_FLAGS "${COMMON_FLAGS} -g -fprofile-arcs -ftest-coverage")
+    set(COMMON_FLAGS "${CMAKE_CXX_FLAGS} -g -fprofile-arcs -ftest-coverage")
```
sas: Why are you doing this?

.travis.yml
```diff
@@ -48,7 +48,7 @@ matrix:
     - os: linux
       env: TARGET="Linux-X86"      LLDB_TESTS="all"  CLANG="0" COVERAGE="1" PLATFORM="1"
     - os: linux
-      env: TARGET="Linux-X86"      LLDB_TESTS="all"  CLANG="0" RELEASE="1"
+      env: TARGET="Linux-X86"      LLDB_TESTS="all"  CLANG="0" COVERAGE="1"  RELEASE="1"
     - os: linux
       env: TARGET="Linux-X86"      LLDB_TESTS="all"  CLANG="1"
```
sas: Why don't we enable coverage here?

.travis.yml
```diff
@@ -57,20 +57,20 @@ matrix:
     - os: linux
       env: TARGET="Linux-X86_64"   LLDB_TESTS="all"  CLANG="0" COVERAGE="1"
```
sas: And here.

.travis.yml
fjricci: For some reason, libprofile for i386 clang doesn't want to install. I can do it locally, but travis can't find the package.

fjricci:
We can't enable coverage always, because I can't get the libprofiler install on x86 to work on Travis.

PR #450, opened 2016-10-24
Fixes for various issues found while trying to run on Yocto Linux

sas:


Sources/Target/Linux/Process.cpp
```diff
@@ -327,13 +327,19 @@ ErrorCode Process::getMemoryRegionInfo(Address const &address,
     uint64_t inode;
     int nread;
 
-    if (std::fgets(buf, sizeof(buf), fp) == nullptr)
-      break;
+    if (::fgets(buf, sizeof(buf), fp) == nullptr) {
```
fjricci: Why are we getting rid of the `std::`?

Sources/Host/Common/Socket.cpp
```diff
@@ -99,11 +99,12 @@ bool Socket::listen(std::string const &address, std::string const &port) {
...
     return false;
 
+  // Disable socket lingering so the process can die quickly when we exit.
   struct linger linger;
-  linger.l_onoff = 1;
+  linger.l_onoff = 0;
   linger.l_linger = 0;
   ::setsockopt(_handle, SOL_SOCKET, SO_LINGER,
-               reinterpret_cast<char *>(&linger), sizeof(linger));
+               reinterpret_cast<void *>(&linger), sizeof(linger));
```
fjricci: I think `::setsockopt` takes a const void *, not sure how much we care

Sources/Host/Common/Socket.cpp
```diff
@@ -135,6 +131,17 @@ bool Socket::listen(std::string const &address, std::string const &port) {
...
 
+  // Disable socket lingering so the process can die quickly when we exit.
+  struct linger linger;
+  linger.l_onoff = 0;
+  linger.l_linger = 0;
+  if (::setsockopt(_handle, SOL_SOCKET, SO_LINGER,
+                   reinterpret_cast<void *>(&linger), sizeof(linger)) == -1) {
+    DS2LOG(Warning,
+           "unable to disable SO_LINGER on the server socket, errno=%s",
+           Stringify::Errno(errno));
```
fjricci: This isn't a posix-specific file

Sources/Target/Linux/Process.cpp
sas: Cause I find it cleaner to use `std::` for things coming from the standard C++ library, and straight up `::` for things coming from the C library. `fgets` for instance is a C function, not a C++ function.

Sources/Host/Common/Socket.cpp
sas: Damn you're right.

Sources/Host/Common/Socket.cpp
```diff
@@ -142,6 +142,16 @@ bool Socket::listen(std::string const &address, std::string const &port) {
            Stringify::Errno(errno));
   }
 
+  // Enable SO_REUSEADDR so we don't crash when trying to reuse a port after a
+  // previous instance of ds2 exits.
+  int enabled = 1;
+  if (::setsockopt(_handle, SOL_SOCKET, SO_REUSEADDR,
```
sas: Yes.

a20012251:
lgtm

Sources/Target/Linux/Thread.cpp
```diff
@@ -105,8 +105,8 @@ ErrorCode Thread::updateStopInfo(int waitStatus) {
     ProcessThreadId ptid(process()->pid(), tid());
     ErrorCode error = process()->ptrace().getSigInfo(ptid, si);
     if (error != kSuccess) {
-      DS2LOG(Warning, "unable to get siginfo_t for tid %d, error=%s", tid(),
-             strerror(errno));
+      DS2LOG(Warning, "unable to get siginfo_t for tid %d, errno=%s", tid(),
+             Stringify::Errno(errno));
```
fjricci: Why is this better? Isn't `strerror` more descriptive than just the errno enum name?

Sources/Target/Linux/Process.cpp
fjricci: This makes it look like ability to drop `std::` is implementation-defined.

http://stackoverflow.com/questions/29897661/when-using-standard-c-functions-in-c-is-the-std-prefix-required

sas: It's not implementation-defined. It supposedly is "deprecated", but we're using `::` everywhere and these two seemed to be outliers so unless we want to change everything, I suppose this way is better.

Sources/Target/Linux/Thread.cpp
sas: Not really. For instance, `ENOENT` would be translated to "No such file or directory" even if we're not using files. Having the errno makes it easier to grep for it in man pages, etc. Also, the formatting with `strerror` looks stupid.

Sources/Host/Common/Socket.cpp
sas: You can pass a pointer as argument to a function that takes a const pointer. The constness of the function argument just indicates that the callee will not modify the data; the pointer you hold being const makes no difference.

PR #459, opened 2016-11-30
Read and write floating point registers on Windows ARM

sas:


sas:
The failing builds are because of the android toolchain script being broken. I'm fixing that in PR #460 and this PR doesn't affect android at all so I'm gonna go ahead and just merge this.

PR #461, opened 2016-12-05
Disable an extra test

sas:
Not sure why this one started failing. It doesn't fail on my host.


sas:
This is useless as well.

PR #462, opened 2017-03-17
Make prepare-android-toolchain.sh work with pre-installed NDKs

sas:


fjricci:
Hmm, what will happen if the pre-installed NDK isn't complete (ie if someone CTRL+C's in the middle of the process)? I ask because I did that earlier. Is there a better way to check than just making sure the directory exists?

fjricci:
Discussed offline, lgtm

PR #466, opened 2017-03-24
Use CMake checks to determine features

sas:
This is more accurate than looking at flags the compiler (or other
toolchain components) might define. In this specific instance, we had a
problem because we were trying to build a ds2 binary for Tizen, using
the Android toolchain, so half of these checks were not accurate.

### DIFF
--- a/Headers/DebugServer2/Host/Linux/ExtraWrappers.h
+++ b/Headers/DebugServer2/Host/Linux/ExtraWrappers.h
@@ -86,21 +86,26 @@ struct xfpregs_struct {
 static inline int posix_openpt(int flags) { return ::open("/dev/ptmx", flags); }
 #endif // !HAVE_POSIX_OPENPT
 
-#if !defined(PLATFORM_ANDROID)
+// Do not use PLATFORM_ANDROID here. We want to test if we're using the Android
+// toolchain, not if we're building for the Android target. In some cases, we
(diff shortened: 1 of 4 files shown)


Headers/DebugServer2/Host/Linux/ExtraWrappers.h
```diff
@@ -86,21 +86,26 @@ struct xfpregs_struct {
...
 
-#if !defined(PLATFORM_ANDROID)
+// Do not use PLATFORM_ANDROID here. We want to test if we're using the Android
+// toolchain, not if we're building for the Android target. In some cases, we
+// build Tizen binaries with the Android toolchain.
+#if !defined(HAVE_GETTID)
 // Android headers do have a wrapper for `gettid`, unlike glibc.
 static inline pid_t gettid() { return ::syscall(SYS_gettid); }
-#endif // !PLATFORM_ANDROID
+#endif // !HAVE_POSIX_OPENPT
```
xiaobai: Small thing: This marks the end of !defined(HAVE_GETTID), not !defined(HAVE_POSIX_OPENPT)

sas: Good catch. I'll fix this.

xiaobai:
LGTM

PR #470, opened 2017-03-31
Add name field in MemoryRegionInfo response

xiaobai:
When LLDB calls GetMemoryRegionInfo on a process that goes through GDBRemoteCommunicationClient, it communicates with ds2 to grab the MemoryRegionInfo from the process. However, in LLDB 4.0, they added a `name` field in the response packet. That is, in addition to `start`, `size`, and `permissions`, `name` is something you can expect to see. I have updated ds2 to reflect this.

### DIFF
--- a/Headers/DebugServer2/Types.h
+++ b/Headers/DebugServer2/Types.h
@@ -301,6 +301,7 @@ struct MemoryRegionInfo {
   Address start;
   uint64_t length;
   uint32_t protection;
+  std::string name;
 #if defined(OS_LINUX)
   std::string backingFile;
   uint64_t backingFileOffset;
@@ -309,17 +310,19 @@ struct MemoryRegionInfo {
 
   MemoryRegionInfo() { clear(); }
   MemoryRegionInfo(Address const &start_, uint64_t length_,
-                   uint64_t protection_) {
+                   uint64_t protection_, std::string name_) {
     clear();
     start = start_;
     length = length_;
     protection = protection_;
+    name = name_;
   }
 
   inline void clear() {
     start.clear();
     length = 0;
     protection = 0;
+    name.clear();
 #if defined(OS_LINUX)
     backingFile.clear();
     backingFileOffset = 0;
--- a/Sources/Target/Linux/Process.cpp
+++ b/Sources/Target/Linux/Process.cpp
@@ -18,6 +18,7 @@
 #include "DebugServer2/Host/Platform.h"
 #include "DebugServer2/Target/Thread.h"
 #include "DebugServer2/Utils/Log.h"
+#include "DebugServer2/Utils/String.h"
 #include "DebugServer2/Utils/Stringify.h"
 
 #include <cerrno>
@@ -339,6 +340,7 @@ ErrorCode Process::getMemoryRegionInfo(Address const &address,
     uint64_t offset;
     unsigned int devMinor, devMajor;
     uint64_t inode;
+    char name[PATH_MAX + 1];
     int nread;
 
     if (::fgets(buf, sizeof(buf), fp) == nullptr) {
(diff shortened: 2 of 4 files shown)


Sources/Target/Linux/Process.cpp
```diff
@@ -363,6 +364,7 @@ ErrorCode Process::getMemoryRegionInfo(Address const &address,
       //
       info.start = last;
       info.length = start - last;
+      info.name = std::string(name);
```
sas: You don't need to explicitly call the `std::string` constructor here I think. You can just do `info.name = name;`.

Sources/Target/Linux/Process.cpp
```diff
@@ -379,6 +381,7 @@ ErrorCode Process::getMemoryRegionInfo(Address const &address,
         info.protection |= ds2::kProtectionExecute;
       while (buf[nread] != '\0' && std::isspace(buf[nread]))
         ++nread;
+      info.name = std::string(name);
```
sas: Same here.

Headers/DebugServer2/Types.h
```diff
@@ -301,6 +301,7 @@ struct MemoryRegionInfo {
   Address start;
   uint64_t length;
   uint32_t protection;
+  std::string name;
```
sas: For the sake of uniformity, you should add a call to `name.clear()` in `MemoryRegionInfo::clear()` and add an `std::string name_` argument to the second constructor.

Sources/Target/Linux/Process.cpp
```diff
@@ -339,6 +339,7 @@ ErrorCode Process::getMemoryRegionInfo(Address const &address,
     uint64_t offset;
     unsigned int devMinor, devMajor;
     uint64_t inode;
+    char name[PATH_MAX];
```
sas: I think by having an array of size `PATH_MAX`, you are *mostly* good, but there are two things I'd do differently:
* `char name[]` should be of size `PATH_MAX + 1` to account for the terminating null character, in case the path is actually of size `PATH_MAX`;
* you should specify the "field size" in your call to `::sscanf()` so you're guaranteed to never go over the size of your array.

Specifying the size in the format string seems to be a PITA because you have to include the numeric value *in* the format string itself, so you have to double-expand `PATH_MAX`...

Looks like there's also a GNU extension that allows you to use the `m` modifier to let sscanf allocate the memory for you, but then you have to `free()` it yourself, and it's not portable:

> • An optional 'm' character. This is used with string conversions (%s, %c, %[), and relieves the caller of the need to allocate a corresponding buffer to hold the input: instead, scanf() allocates a buffer of sufficient size, and assigns the address of this buffer to the corresponding pointer argument, which should be a pointer to a char * variable (this variable does not need to be initialized before the call). The caller should subsequently free(3) this buffer when it is no longer required.

😢

Sources/Target/Linux/Process.cpp
fjricci: Do you need the explicit conversion?

Headers/DebugServer2/Types.h
xiaobai: Done.

Sources/Target/Linux/Process.cpp
```diff
@@ -37,6 +37,9 @@ using ds2::Utils::Stringify;
 
 #define super ds2::Target::POSIX::ELFProcess
 
+#define STR_HELPER(s) #s
```
sas: Argument to macros should be in capital letters, and you should put this in a shared location. `Headers/DebugServer2/Utils/String.h` seems like a good candidate.

xiaobai: Resolved

PR #471, opened 2017-03-31
Use android NDK r14b

sas:
Up from r13b.


xiaobai:
LGTM

sas:
This exposes a bug in the android tests so unfortunately we won't be able to upgrade for now.

PR #472, opened 2017-03-31
Fix Windows AppVeyor build to work with update to MemoryRegionInfo

xiaobai:
Changing MemoryRegionInfo broke Windows builds. For now, we can put an empty string until we can figure out what to properly put here.

### DIFF
--- a/Sources/Target/Windows/Process.cpp
+++ b/Sources/Target/Windows/Process.cpp
@@ -368,14 +368,15 @@ ErrorCode Process::getMemoryRegionInfoInternal(Address const &address,
       _handle, reinterpret_cast<LPVOID>(address.value()),
       reinterpret_cast<PMEMORY_BASIC_INFORMATION>(&mem), sizeof(mem));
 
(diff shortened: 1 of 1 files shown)


Sources/Target/Windows/Process.cpp
```diff
@@ -368,9 +368,10 @@ ErrorCode Process::getMemoryRegionInfoInternal(Address const &address,
       _handle, reinterpret_cast<LPVOID>(address.value()),
       reinterpret_cast<PMEMORY_BASIC_INFORMATION>(&mem), sizeof(mem));
 
+  //TODO: Empty string is a work-around. Fix this at some point.
```
sas: Nit: `// TODO` instead of `//TODO`.

sas: I think the style checker will complain.

fjricci:
Looks like this didn't pass travis builds.

sas:
For some reason there is no travis/appveyor run at all in this PR...

PR #473, opened 2017-04-03
Test aarch64

fjricci:
Probably will be a bit before this works fully, but might as well start trying.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ matrix:
     - os: linux
       env: TARGET="Android-ARM" ANDROID_PLATFORM="16"
     - os: linux
-      env: TARGET="Android-ARM64"
+      env: TARGET="Android-ARM64" LLDB_TESTS="all" PLATFORM="1" COVERAGE="1"
     - os: linux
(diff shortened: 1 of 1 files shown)


Support/Scripts/install-android-emulator.sh
```diff
@@ -41,7 +41,15 @@ if [ ! -f "$android_script" ]; then
...
-package_list+=("android-${api_level}", "sys-img-${emulator_image_arch}-android-${api_level}")
+# This is required due to dependency issues in the sdk. Hopefully this can be removed in future
+# versions of the sdk.
+if [ "${api_level}" -gt 23 ]; then
+  package_list=("android-23", "platform-tools", "tools", "android-${api_level}")
+else
+  package_list=("android-${api_level}")
+fi
+
+package_list+=("sys-img-${emulator_image_arch}-android-${api_level}")
```
sas: Why not put `android-${api_level}` here?

fjricci: actually that's not a bad idea

sas:
This is stuck with emulator issues for now so probably won't go in in the near future.

PR #474, opened 2017-04-03
Fix Travis build failures introduced by memory region packet changes

fjricci:
lol@lldbtestsuite

### DIFF
--- a/Support/Testing/Patches/memory-region-info.patch
+++ b/Support/Testing/Patches/memory-region-info.patch
@@ -1,0 +1,13 @@
+diff --git a/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py b/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py
+index d63ddbe..68f4f50 100644
+--- a/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py
++++ b/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py
+@@ -698,7 +698,7 @@ class GdbRemoteTestCaseBase(TestBase):
+ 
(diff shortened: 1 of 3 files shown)


Support/Testing/Patches/memory-region-info.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py b/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py
+index d63ddbe..68f4f50 100644
+--- a/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py
++++ b/packages/Python/lldbsuite/test/tools/lldb-server/gdbremote_testcase.py
+@@ -698,7 +698,7 @@ class GdbRemoteTestCaseBase(TestBase):
+ 
+         # Validate keys are known.
+         for (key, val) in list(mem_region_dict.items()):
+-            self.assertTrue(key in ["start", "size", "permissions", "error"])
++            self.assertTrue(key in ["start", "size", "permissions", "error", "name"])
```
sas: This is funny. This was added because upstream needs it now. I think when we update to 4.0 test suite, you'll have to revert this change.

fjricci: Well I mean, presumably since upstream needs it, they'll want this patch. Actually how is upstream passing tests without this...

fjricci: Oh wait upstream added this. I'll cherry-pick their patch instead.

fjricci: nvm, I can't do that, it's too big of a patch. I'll just keep this around until we update and get a merge conflict.

PR #476, opened 2017-04-05
Fixes for android 24 emulator

fjricci:


Support/Scripts/install-android-emulator.sh
```diff
@@ -32,7 +32,7 @@ package_list=()
 
 # This is required due to dependency issues in the sdk. Hopefully this can be removed in future
 # versions of the sdk.
-if [ "${api_level}" -gt 23 ]; then
```
sas: I'm not sure you should remove these quotes. If the argument is actually a numeric argument, it won't make a difference, but if the argument is not (error case), the quotes will prevent `[` from failing with stupid errors.

Support/Testing/Travis/before-script.sh
```diff
@@ -20,7 +20,7 @@ if [[ "$TARGET" == Android-* && -n "${LLDB_TESTS-}" ]]; then
 
   case "${TARGET}" in
     "Android-ARM") emulator_arch="arm";;
-    "Android-ARM64") emulator_arch="aarch64";;
+    "Android-ARM64") emulator_arch="ranchu-arm64";;
```
sas: This is a little bit pathetic :D

Support/Scripts/install-android-emulator.sh
fjricci: I'm pretty sure this didn't work with the quotes.

sas: I'm pretty sure language-wise it'll make no difference *unless* `${api_level}` contains something with spaces.

fjricci: The language spec says you're right. I'll double check that it works tomorrow.

Support/Testing/Travis/before-script.sh
xiaobai: Ranchu is the machine type used when invoking qemu-android-aarch64 or whatever it's called. I wish it was just `emulator64-arm64` with the default qemu argument being `-machine type=ranchu`. Including it in the name of the emulator binary feels strange. :/

PR #479, opened 2017-04-18
Changing method of finding DT_DEBUG

xiaobai:
This is a step towards refactoring generalized ELF Process functionality. Getting the address containing shared library info is kind of a complicated thing, and I'm not sure there's any 100% foolproof way of doing it easily (I hope I eat my words). Specifically, here is how it works:

First: In an ELF file, we have a program header (PHDR) used by the dynamic linker to figure out how to actually build a process image from the file. The ELF specification says this is optional, but when you have an executable it is necessary. We can expect it to exist for things we want ds2 to attach to (I think). Next, we recognize that the ELF header is no bigger than 64 bytes (ELF32 is 52 bytes). This is convenient, as in practice, the PHDR is usually right after the header. It is possible that it will not be, but I think it's perfectly reasonable to assume it will be. Therefore, I believe it is possible to get the image's load base by getting the PHDR and subtracting the size of the ELF header.

Second: When searching for the file's load base, it is not enough to find the first PT_LOAD and call it a day. It is valid for a linker to rearrange the PT_LOADs however it wants, so we look at every PT_LOAD and find the smallest p_vaddr associated with all PT_LOAD segments. In practice, it is usually the first one. I don't believe this has any significant performance issues, so better to be safe than sorry. Additionally, the PT_DYNAMIC segment can come before any PT_LOAD so it's not enough to stop when you find the PT_DYNAMIC segment (if it exists).

I have tested this on Linux-x86_64 with both gold and LLD, and both work correctly. I have also tested this change on Android ARM (API 22). It works correctly with gold, but has trouble with LLD. I believe this to be an LLD issue, as I detected DT_DEBUG in the binary built with gold but not in the one built with LLD. Android ARM still will run the binary correctly, albeit slower as it must send a large amount of memory reads.

A few tidbits about the last small change:

The call to updateInfo in ELFProcess::getSharedLibraryInfoAddress was simply a way to call updateAuxiliaryVector and set the loadBase and entryPoint correctly. However, loadBase will soon be defunct and a lot of code in ELFProcess::updateInfo will be dead code. I felt it was more appropriate here to call updateAuxiliaryVector directly here.

Future work:
ELFProcess::updateInfo should be refactored and have the dead code removed. We should also examine derived classes (e.g. Linux::Process) to see if the changes made in ELFProcess might pose any future issues to derived classes in general, as they may rely on behavior of ELFProcess. One example is Linux::Process::updateInfo calls super::updateinfo (which is ELFProcess::updateInfo) only to call updateAuxiliaryVector(). There may be more examples.

### DIFF
--- a/Sources/Target/POSIX/ELFProcess.cpp
+++ b/Sources/Target/POSIX/ELFProcess.cpp
@@ -53,52 +53,66 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
   PHDR ph;
   DYNSYM dynsym;
 
-  CHK(process->readMemory(process->loadBase(), &eh, sizeof(eh)));
-
-  if (eh.e_phnum == 0) {
+  // 1. Find the PHDR, number of entries, size of each entry, and the loadBase
+  //    of the binary.
+  //    Assumption: PHDR is attached to the ELF header. This isn't enforced but
+  //    it almost always occurs in practice.
+  uint64_t phdrAddress = process->getAuxiliaryVectorValue(AT_PHDR);
+  size_t phdrEntrySize = process->getAuxiliaryVectorValue(AT_PHENT);
+  size_t phdrNumEntry = process->getAuxiliaryVectorValue(AT_PHNUM);
+  uint64_t loadBase = phdrAddress - sizeof(eh);
+  if (phdrAddress == 0 || phdrEntrySize == 0 || phdrNumEntry == 0) {
+    DS2LOG(Debug, "Unable to locate PHDR info, "
+                  "Dynamic Library info will not be available.");
     return kErrorUnsupported;
   }
 
-  // 1. Find PT_DYNAMIC (.dynamic) program header, and the
-  //    expected load base of this ELF.
+  // 2. Find PT_DYNAMIC (.dynamic) program header and what the ELF believes is
+  //    the loadBase. Find the size of the dynamic section as well.
   uint64_t elfLoadBase = 0;
   bool foundElfLoadBase = false;
-  uint64_t phdrAddress = process->loadBase() + eh.e_phoff;
-  for (size_t n = 0; n < eh.e_phnum; n++) {
-    CHK(process->readMemory(phdrAddress, &ph, sizeof(ph)));
+  uint64_t dotDynamicAddress = 0;
+  bool foundDotDynamicAddress = false;
+  uint64_t dotDynamicSize = 0;
+  for (size_t n = 0; n < phdrNumEntry; n++) {
+    CHK(process->readMemory(phdrAddress + n * phdrEntrySize, &ph, sizeof(ph)));
 
-    if (!foundElfLoadBase && ph.p_type == PT_LOAD) {
-      elfLoadBase = ph.p_paddr;
+    if ((!foundElfLoadBase || ph.p_vaddr < elfLoadBase) &&
+        ph.p_type == PT_LOAD) {
+      elfLoadBase = ph.p_vaddr;
       foundElfLoadBase = true;
     }
 
     if (ph.p_type == PT_DYNAMIC) {
-      break;
+      dotDynamicAddress = ph.p_vaddr;
+      dotDynamicSize = ph.p_memsz;
+      foundDotDynamicAddress = true;
     }
-
-    phdrAddress += eh.e_phentsize;
   }
 
-  if (ph.p_type != PT_DYNAMIC) {
+  if (!foundDotDynamicAddress || !foundElfLoadBase) {
+    DS2LOG(Debug, "Unable to locate PT_DYNAMIC or any PT_LOAD, "
+                  "Dynamic Library info will not be available.");
     return kErrorUnsupported;
   }
 
-  // 2. Find DT_DEBUG in .dynamic section
-  uint64_t dynsymAddress = process->loadBase() + ph.p_paddr - elfLoadBase;
-  size_t entriesCount = ph.p_memsz / sizeof(dynsym);
+  // 3. Find DT_DEBUG in .dynamic section
+  dotDynamicAddress += (loadBase - elfLoadBase);
+  size_t entriesCount = dotDynamicSize / sizeof(dynsym);
   for (size_t n = 0; n < entriesCount; n++) {
-    CHK(process->readMemory(dynsymAddress, &dynsym, sizeof(dynsym)));
+    CHK(process->readMemory(dotDynamicAddress + n * sizeof(dynsym), &dynsym,
+                            sizeof(dynsym)));
 
-    // 3. Return &DT_DEBUG->d_ptr, which is filled by the dynamic linker
+    // 4. Return &DT_DEBUG->d_ptr, which is filled by the dynamic linker
     //    pointing to the link_map.
     if (dynsym.d_tag == DT_DEBUG) {
-      address = dynsymAddress + sizeof(dynsym.d_tag);
+      address = (dotDynamicAddress + n * sizeof(dynsym)) + sizeof(dynsym.d_tag);
       return kSuccess;
     }
-
-    dynsymAddress += sizeof(dynsym);
   }
 
+  DS2LOG(Debug, "DT_DEBUG was not found in .dynamic, "
+                "Dynamic Library info will not be available.");
   return kErrorUnsupported;
 }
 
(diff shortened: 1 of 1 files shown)


xiaobai:
Just checked, this works on android-x86 with gold, but fails with LLD. I believe this to be an problem with LLDB, but it warrants further investigation.

Also, I updated the PR to fix some code style issues found on Travis.

xiaobai:
Found a bug, amended my commit.

xiaobai:
After fixing the bug (erroneously using ph.p_memsz instead of using PT_DYNAMIC's memsz), here is what I've tested:

Linux-x86_64 w/ gold: Works
Linux-x86_64 w/ LLD: Works
android-arm w/ gold: Works
android-arm w/ LLD: Works
android-x86 w/ gold: Works
android-x86 w/ LLD: Fails (LLDB tries to read something at location 0x0). I believe this to be unrelated.

I'm believe the PR is ready for review whenever the reviewers are ready. :D

xiaobai:
I updated the commit to delete dead code and add/update comments for clarity.

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -475,9 +477,7 @@ ErrorCode ELFProcess::getSharedLibraryInfoAddress(Address &address) {
     //
     // Force the updating of process information.
     //
-    _info.pid = -1;
-
-    ErrorCode error = updateInfo();
+    ErrorCode error = updateAuxiliaryVector();
```
sas: I like this. :p

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -298,26 +311,15 @@ ErrorCode ELFProcess::updateInfo() {
...
-    _loadBase = prevMri.start;
+    // Here, we use the interpreter's base address
+    // instead of the binary's. The interpreter
+    // has an ELF header just the same as the
+    // binary, so we can get most of the necessary
+    // info from it.
+    // If this fails for any reason, an easy fallback
+    // is to grab the PHDR and subtract the size of
+    // the ELF header. This should be good enough, though.
+    _loadBase = getAuxiliaryVectorValue(AT_BASE);
```
sas: Do we still need all of this? Why don't we just get rid of `_loadBase` altogether?

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,43 +53,56 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
   PHDR ph;
   DYNSYM dynsym;
 
-  CHK(process->readMemory(process->loadBase(), &eh, sizeof(eh)));
-
-  if (eh.e_phnum == 0) {
+  // 1. Find the PHDR, number of entries, size of
```
sas: For your comments, you can use the full 80 columns. It seems like you set your editor to something like 65 columns or so.

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,43 +53,56 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
...
+  //    each entry, and the loadBase of the binary.
+  //    Assumption: PHDR is attached to the ELF
+  //    header. This isn't enforced but it almost
+  //    always occurs in practice.
+  uint64_t phdrAddress = process->getAuxiliaryVectorValue(AT_PHDR);
+  size_t phdrEntrySize = process->getAuxiliaryVectorValue(AT_PHENT);
+  size_t phdrNumEntry = process->getAuxiliaryVectorValue(AT_PHNUM);
+  uint64_t loadBase = phdrAddress - sizeof(eh);
```
sas: Clever.

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,43 +53,56 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
...
+  if (phdrAddress == 0 || phdrEntrySize == 0 || phdrNumEntry == 0) {
```
sas: I wouldn't mind having a `Debug` log here if one of these checks fails, to indicate that dynamic library info won't be available.

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,43 +53,56 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
...
-  //    expected load base of this ELF.
+  // 2. Find PT_DYNAMIC (.dynamic) program header
+  //    and what the ELF believes is the loadBase.
+  //    Find the size of the dynamic section as well.
   uint64_t elfLoadBase = 0;
   bool foundElfLoadBase = false;
-  uint64_t phdrAddress = process->loadBase() + eh.e_phoff;
-  for (size_t n = 0; n < eh.e_phnum; n++) {
+  uint64_t dynamicAddress = 0;
+  uint64_t dynamicMemsz = 0;
```
sas: Nit: These variable could be named `dotDynamicAddress` and `dotDynamicMemSize` to be more explicit.

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,43 +53,56 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
...
 
     if (ph.p_type == PT_DYNAMIC) {
-      break;
+      dynamicAddress = ph.p_vaddr;
+      dynamicMemsz = ph.p_memsz;
+      foundDynamicAddress = true;
     }
 
-    phdrAddress += eh.e_phentsize;
+    phdrAddress += phdrEntrySize;
```
sas: Nit: Not your code, but I think changing the value of `phdrAddress` makes the code slightly harder to follow. Doing something like

    CHK(process->readMemory(phdrAddress + n * sizeof(ph), &ph, sizeof(ph)));

seems easier to read for me.

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,43 +53,56 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
...
   }
 
-  if (ph.p_type != PT_DYNAMIC) {
+  if (!foundDynamicAddress || !foundElfLoadBase) {
```
sas: Same thing here about having a `Debug` log.

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,43 +53,56 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
...
     return kErrorUnsupported;
   }
 
-  // 2. Find DT_DEBUG in .dynamic section
-  uint64_t dynsymAddress = process->loadBase() + ph.p_paddr - elfLoadBase;
-  size_t entriesCount = ph.p_memsz / sizeof(dynsym);
+  // 3. Find DT_DEBUG in .dynamic section
+  uint64_t dynsymAddress = loadBase - elfLoadBase + dynamicAddress;
```
sas: Again, not your code, but I have no idea why this variable is named `dynsymAddress`. `.dynsym` is a different section with a different purpose. This is confusing.

I wouldn't mind reusing `dynamicAddress` here and just adjusting it with the load offset.

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: Haha let me reformat the comments.

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: :)

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: Agreed, I'll go ahead and modify the commit to include one.

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: I like this idea! I'm bad at naming things. :(

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: I agree. Will amend the commit.

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: Will amend.

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: I'll change dynamicAddress to dotDynamicAddress and reuse it here as well, per your suggestions.

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: Currently, ELFProcess::updateInfo is used as a fallback for when platform-specific implementations of updateInfo fail. As we talked about in person, this function should fill in as much info as possible and leave platform-specific implementations of updateInfo to fill in platform-specific information. I'll change this in a separate commit and submit a pull request when I get the time.

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: :D

Sources/Target/POSIX/ELFProcess.cpp
xiaobai: We don't need _loadBase at all. We can use the PHDR to find the ELF header of the binary (the same way the modified function does things) and then start extracting info that way. We can also forego the ELF header validation, since if auxv exists, then we have an ELF. to begin with.

xiaobai:
I've updated the commit per @sas's suggestions. :D

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -53,52 +53,66 @@ inline ErrorCode GetELFSharedLibraryInfoAddress(ELFProcess *process,
...
     if (dynsym.d_tag == DT_DEBUG) {
-      address = dynsymAddress + sizeof(dynsym.d_tag);
+      address = (dotDynamicAddress + n * sizeof(dynsym)) + sizeof(dynsym.d_tag);
       return kSuccess;
     }
 
-    dynsymAddress += sizeof(dynsym);
   }
 
+  DS2LOG(Debug, "DT_DEBUG was not found in .dynamic");
```
sas: Same here, just one log message is better.

    DS2LOG(Debug, "DT_DEBUG was not found in .dynamic, "
                  "dynamic Library info will not be available.");

PR #482, opened 2017-05-01
Ensure that breakpoints can be enabled successfully before adding

fjricci:


Sources/Core/BreakpointManager.cpp
```diff
@@ -49,6 +49,15 @@ ErrorCode BreakpointManager::add(Address const &address, Type type, size_t size,
     // the newly added breakpoint too.
     if (enabled()) {
       return enableLocation(site);
+    } else {
+      ErrorCode isValid = enableLocation(site);
+      disableLocation(site);
```
sas: I think you should `disableLocation(site)` only if the `enableLocation()` at the previous line returned `kSuccess`.

Sources/Core/BreakpointManager.cpp
```diff
@@ -49,6 +49,15 @@ ErrorCode BreakpointManager::add(Address const &address, Type type, size_t size,
...
+
+      if (isValid != kSuccess) {
+        remove(address);
+      }
+
+      return isValid;
     }
   }
```
sas: You can remove the `return kSuccess;` at the end of this function now that it's dead code.

fjricci: It's not dead code if the breakpoint already exists.

PR #487, opened 2017-05-05
Disable nondebugserver tests

fjricci:
This will hopefully speed up our builds, and these tests don't involve ds2 anyway.


fjricci:
This currently depends on https://github.com/facebook/ds2/pull/480

PR #491, opened 2017-05-10
Disable some flaky tests

fjricci:
Let's leave this open for a bit so we can add more to it if we see more flakes.


fjricci:
Things seem to be pretty stable in the CRON jobs, so we may not even need this.

fjricci:
Ok just kidding, as soon as I say that, a CRON job failed.

PR #493, opened 2017-06-20
Implement memory allocation for aarch64 Linux

sas:


Headers/DebugServer2/Utils/HexValues.h
```diff
@@ -42,7 +42,7 @@ template <typename T> static inline std::string ToHex(T const &vec) {
   std::string result;
   for (char n : vec) {
     result += NibbleToHex(n >> 4);
-    result += NibbleToHex(n & 15);
+    result += NibbleToHex(n & 0x0f);
```
a20012251: It'd be nice to define a uint4_t with static casts :3

Sources/Host/Linux/ARM64/PTraceARM64.cpp
```diff
@@ -60,20 +60,11 @@ int PTrace::getMaxHardwareWatchpoints(ProcessThreadId const &ptid) {
 ErrorCode PTrace::readCPUState(ProcessThreadId const &ptid,
                                ProcessInfo const &pinfo,
                                Architecture::CPUState &state) {
-  pid_t pid;
-
-  ErrorCode error = ptidToPid(ptid, pid);
```
a20012251: probably useful at some point

xiaobai: I figured. It's weird that it was left in if it wasn't being used, though.

sas: Yeah I think that's when I was doing the initial support for aarch64, I used the ARM version as a template and never realized that the pid wasn't used in the ARM64 version.

Headers/DebugServer2/Utils/HexValues.h
sas: I think the goal here is to just be a little bit more explicit about what we're doing. Having a uint4_t would be cute but I'm not sure I want to implement that right now :p

Headers/DebugServer2/Architecture/ARM64/CPUState.h
```diff
@@ -38,10 +38,10 @@ typedef ds2::Architecture::ARM::CPUState CPUState32;
 //
 struct CPUState64 {
   union {
-    uint64_t regs[32 + 1 + 1];
+    uint64_t regs[31 + 1 + 1 + 1];
```
sas: That's just a semi-useful way of keeping track of where these registers come from. There are 31 GP registers (x0...x30) and sp, pc, and cpsr. This change doesn't have any actual impact on how the code runs.

xiaobai: 👍

PR #498, opened 2017-07-28
Bump cmake version to 3.8

xiaobai:
Let's see if this fixes the problem with incorrect ninja version detection. If not, we'll switch to Make instead of Ninja for now.


xiaobai:
After #499 we shouldn't need this. I still see this as a valuable change, however. What do y'all think? :p

xiaobai:
3.7 probably works just fine. If we ever need to upgrade, I'll submit another PR.

PR #500, opened 2017-07-30
Remove temporary zips and tarballs during setup for more space

xiaobai:
Currently one of the targets fails pretty consistently with the error `NAND: could not write file /home/travis/.android/avd/test.avd/userdata-qemu.img, No space left on device`. In this PR, I aim to try and reduce the amount of space as much as possible. Currently all I've done is delete the sdk's zip/tarball. If this isn't enough, I can stack more diffs deleting more things.

### DIFF
--- a/Support/Scripts/install-android-emulator.sh
+++ b/Support/Scripts/install-android-emulator.sh
@@ -51,6 +51,7 @@ if [ ! -f "$android_script" ]; then
     "tgz") tar -C /tmp -xvf "/tmp/${android_sdk_package}";;
     *) die "Unknown archive extension '${package_extension}'."
   esac
+  rm -f "/tmp/${android_sdk_package}"
 fi
 
 for package in "${package_list[@]}"; do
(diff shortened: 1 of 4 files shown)


Support/Scripts/install-android-emulator.sh
```diff
@@ -51,6 +51,7 @@ if [ ! -f "$android_script" ]; then
     "tgz") tar -C /tmp -xvf "/tmp/${android_sdk_package}";;
     *) die "Unknown archive extension '${package_extension}'."
   esac
+  rm -rf "/tmp/${android_sdk_package}"
```
sas: Nit: the `-r` isn't required.

Support/Scripts/run-gdb-tests.sh
```diff
@@ -27,6 +27,7 @@ gdb_src_path="$build_dir/$gdb_basename"
 
 wget "$gdb_ftp_path/$gdb_basename.tar.gz"
 tar -zxf "$gdb_basename.tar.gz"
+rm -r "${gdb_basename}.tar.gz"
```
sas: `s/-r/-f/`

xiaobai: good catch :D

PR #501, opened 2017-07-31
Inspect what takes up so much space on Travis

xiaobai:
This is stupid. For some reason, Travis runs out of space even if we remove a lot of the temporary files we download (*.zip and *.tar.gz files). See #500 for more context. What's taking up so much space, and why? Let's find out.


xiaobai:
This PR accomplished its goal. :D

PR #504, opened 2017-08-01
Remove perl5 and otp installs on travis

xiaobai:
This is shitty hack. I feel bad about doing this, but I think it should fix our things in the short term while we come up with a better solution for the long term. I'm thinking we can probably play around with Travis settings until we find a sweet spot, but that will take more time to read the documentation and engage the Travis community.

### DIFF
--- a/Support/Testing/Travis/script.sh
+++ b/Support/Testing/Travis/script.sh
@@ -14,6 +14,8 @@ source "$top/Support/Scripts/common.sh"
 
 cformat="clang-format-4.0"
 
+"$top/Support/Testing/Travis/travis-hacks.sh"
+
 # If we're at the root of the repository, create a build directory and go
 # there; otherwise, assume that we're already in the build directory the user
 # wants to use.
--- a/Support/Testing/Travis/travis-hacks.sh
+++ b/Support/Testing/Travis/travis-hacks.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+##
+## Copyright (c) 2014-present, Facebook, Inc.
+## All right reserved.
+##
+## This source code is licensed under the University of Illinois/NCSA Open
+## Source License found in the LICENSE file in the root directory of this
+## source tree. An additional grant of patent rights can be found in the
+## PATENTS file in the same directory.
+
+set -eu
+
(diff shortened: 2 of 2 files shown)


Support/Testing/Travis/script.sh
```diff
@@ -14,6 +14,17 @@ source "$top/Support/Scripts/common.sh"
 
 cformat="clang-format-4.0"
 
+# We're running out of disk space on Travis. This is a dirty, dirty hack. We shouldn't have to do this.
+# In the home directory, /home/travis/perl5 takes up roughly 1.1GB of space and /home/travis/otp takes
+# up roughly 3.0GB. We should be able to recover this space on travis by removing these things, since
+# we never use them.
+if [ -e "~/perl5" ]; then
```
sas: There's no need to check if the folder exists because you're using `rm -f` anyway.

Support/Testing/Travis/script.sh
```diff
@@ -14,6 +14,17 @@ source "$top/Support/Scripts/common.sh"
...
+  rm -rf ~/perl5
+fi
+if [ -e "~/otp" ]; then
```
sas: Ditto.

Support/Testing/Travis/script.sh
```diff
@@ -14,6 +14,17 @@ source "$top/Support/Scripts/common.sh"
 
 cformat="clang-format-4.0"
 
+# We're running out of disk space on Travis. This is a dirty, dirty hack. We shouldn't have to do this.
```
sas: Are you sure these lines are less than 80 columns?

Support/Testing/Travis/travis-hacks.sh
```diff
@@ -0,0 +1,17 @@
...
+## All right reserved.
+##
+## This source code is licensed under the University of Illinois/NCSA Open
+## Source License found in the LICENSE file in the root directory of this
+## source tree. An additional grant of patent rights can be found in the
+## PATENTS file in the same directory.
+
+set -eu
+
+# We're running out of disk space on Travis. This is a dirty hack. We shouldn't have to do this.
```
sas: 80 columns

PR #505, opened 2017-08-01
Rewrite scripts for testing on Android

xiaobai:
I'm going to go ahead and submit this now. I'm curious if the newer method of installing an install emulator somehow takes up less space or can create a smaller emulator image for android-x86. Additionally, this should lay the groundwork for being able to test on android arm64.

I'm hoping this can somehow help us reduce disk usage so we can get android-x86 testing properly again and I won't have to merge #504.

### DIFF
--- a/Support/Testing/Travis/before-script.sh
+++ b/Support/Testing/Travis/before-script.sh
@@ -15,18 +15,19 @@ source "$top/Support/Scripts/common.sh"
 if [[ "$TARGET" == Android-* && -n "${LLDB_TESTS-}" ]]; then
   case "$(uname)" in
     "Linux")  platform_name="linux";;
-    "Darwin") platform_name="macosx";;
+    "Darwin") platform_name="darwin";;
   esac
 
   case "${TARGET}" in
-    "Android-ARM") emulator_arch="arm";;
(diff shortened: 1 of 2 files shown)


Support/Testing/Travis/before-script.sh
```diff
@@ -15,18 +15,18 @@ source "$top/Support/Scripts/common.sh"
...
-    "Darwin") platform_name="macosx";;
+    "Darwin") platform_name="darwin";;
   esac
 
   case "${TARGET}" in
-    "Android-ARM") emulator_arch="arm";;
-    "Android-ARM64") emulator_arch="ranchu-arm64";;
-    "Android-X86") emulator_arch="x86";;
+    "Android-ARM") emulator_arch="emulator64-arm";;
+    "Android-ARM64") emulator_arch="emulator";;
```
sas: Is this correct?

xiaobai: Yes, I believe this is correct.

xiaobai:
There are a lot of issues here. Installing qt doesn't help, because the emulator still complains about the libraries installed. I think we should explore other options.

xiaobai:
I'm going to try to make this work.

PR #508, opened 2017-08-15
Ignore .swp files

xiaobai:
I often have files open in vim when I am looking at the status of the branch I'm on or staging files to commit. Seeing the .swp files there when i run `git status` just adds clutter. Let's ignore them, as they serve no benefit to actually have.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build/
 build-*/
 html/
+*.swp


sas:
👍 

I was curious why I never saw these files in my git status, and realized it's because I have `.*.swp` in my `~/.gitignore`. :p

PR #510, opened 2017-08-15
Support read-only hardware watchpoints on x86

AndrewSB:


Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -94,13 +104,57 @@ ErrorCode HardwareBreakpointManager::disableLocation(Site const &site,
...
+  CHK(_process->readMemory(address, &value, sizeof(value)));
+
+  // store the new value at address in _locations
+  auto loc = std::find_if(_locations.begin(), _locations.end(),
+                          [address](const Location &l){
+                            return l.address == address;
+                          });
+  if (loc == _locations.end()) {
+     // exception. It isn't  possible to hit at an address that isn't enabled
+    return kErrorInvalidAddress;
```
AndrewSB: not sure if this is the right error to return

Headers/DebugServer2/Core/HardwareBreakpointManager.h
```diff
@@ -67,7 +67,13 @@ class HardwareBreakpointManager : public BreakpointManager {
                                 std::vector<uint64_t> &regs) const;
 
 protected:
-  std::vector<uint64_t> _locations;
+  struct Location {
+    Location()
+      :address(0), value(0){}
```
fjricci: You'll need to run clang-format on this. You can either:
1) Install and run clang-format manually (I don't think this works on devservers)
2) Look at the Travis "STYLE" failures and manually apply
3) install clang-format and run: TARGET=Style ./Support/Testing/Travis/script.sh (see previous comment about no devservers, this will probably require temporarily changing script.sh to use an unversioned clang-format)

Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -40,9 +39,12 @@ ErrorCode HardwareBreakpointManager::add(Address const &address, Type type,
 }
 
 ErrorCode HardwareBreakpointManager::remove(Address const &address) {
-  auto loc = std::find(_locations.begin(), _locations.end(), address);
+  auto loc = std::find_if(_locations.begin(), _locations.end(),
+                          [address](const Location &l){
+                            return l.address == address;
+                          });
```
fjricci: Please factor out the comparator and pass around as a function pointer.

Sources/Core/ARM/HardwareBreakpointManager.cpp
```diff
@@ -35,6 +35,12 @@ size_t HardwareBreakpointManager::maxWatchpoints() {
   return _process->getMaxWatchpoints();
 }
 
+bool HardwareBreakpointManager::checkIfWrittenTo(Address const &address) {
+  bool implemented = false;
+  DS2ASSERT(implemented);
+  return implemented;
```
fjricci: You can just put a DS2BUG("not implemented") instead

Sources/Core/ARM/HardwareBreakpointManager.cpp
```diff
@@ -35,6 +35,12 @@ size_t HardwareBreakpointManager::maxWatchpoints() {
   return _process->getMaxWatchpoints();
 }
 
+bool HardwareBreakpointManager::checkIfWrittenTo(Address const &address) {
```
fjricci: wasWritten()

Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -102,6 +104,44 @@ ErrorCode HardwareBreakpointManager::disableLocation(Site const &site,
   return error;
 }
 
+// workaround for hardware readonly watchpoints not being implememted in
+// hardware ¯\_(ツ)_/¯ https://github.com/facebook/ds2/issues/257
```
fjricci: I don't think this really qualifies as a workaround, since the lack of hardware implementation is an intentional feature. You could just say that this is a software implementation of read-only watchpoints or something.

Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -102,6 +104,44 @@ ErrorCode HardwareBreakpointManager::disableLocation(Site const &site,
...
+bool HardwareBreakpointManager::hardwareReadonlyWorkaround(
```
fjricci: You can probably just inline this function.

Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -102,6 +104,44 @@ ErrorCode HardwareBreakpointManager::disableLocation(Site const &site,
...
+}
+
+void HardwareBreakpointManager::storeNewValueAtAddress(Address const &address) {
+  uint64_t newValueAtAddress = 0; // TODO(asb) get the value from ptrace
+
+  // store the new value at address in _locations
+  auto loc = std::find_if(_locations.begin(), _locations.end(),
+                          [address](const Location &l){
+                            return l.address == address;
+                          });
```
fjricci: see previous comment about factoring this out.

Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -113,7 +153,7 @@ int HardwareBreakpointManager::getAvailableLocation() {
                               return l.address == 0;
                             });
   DS2ASSERT(it != _locations.end());
-  
+
```
fjricci: NIT: remove accidental whitespace change

Sources/Core/X86/HardwareBreakpointManager.cpp
```diff
@@ -32,6 +32,27 @@ size_t HardwareBreakpointManager::maxWatchpoints() {
   return 4; // dr0, dr1, dr2, dr3
 }
 
+bool HardwareBreakpointManager::checkIfWrittenTo(Address const &address) {
```
fjricci: wasWritten()

Sources/Core/HardwareBreakpointManager.cpp
fjricci: I'm fine with a DS2ASSERT(loc != _locations.end());

Sources/Core/HardwareBreakpointManager.cpp
AndrewSB: you mean define a function like
```
inline bool addressInMap(const HardwareBreakpointManager::Location &location,
                         const u_int64_t address)
                         __attribute__((always_inline)) {
  return location.address == address;
}
```

and use it everywhere like 

```
std::find_if(_locations.begin(), _locations.end(),
    [address](const Location &location) { return addressInMap(location, address); });
```
?

fjricci: Actually, I have a different idea. We could just use a std::map for `_locations`, and remove the `DS2ASSERT(_locations.size() == maxWatchpoints());` (since you can't fix the size of a map). we already check `_sites.size() >= maxWatchpoints()` in `HardwareBreakpointManager::add`, so there's really no reason to also restrict the size of `_locations`. If you use <address, value> as <key, value> in a map, you don't need the find_if's anymore.

AndrewSB: thats great, I'll do that in a commit in the next hour or so :)

AndrewSB:
@fjricci should I squash all/some of my commits?

fjricci:
Stacked commits are good (much easier to review and understand than one big blob), but they should be stacked by functionality and not by time (so a commit like `fix style` or `address comments` should be squashed). Each point in the stack should have correct style and be a logical step toward the end result. For example, we don't want a commit to add `value` to _locations and then another commit to remove it, but it's fine/good to add one level of functionality at a time.

AndrewSB:
sounds good! I'll rebase functionally 😄

fjricci:
Pro tip: use `git rebase --ignore-date` on your local branch to make the commits show up in the correct order here after squashing.

AndrewSB:
@fjricci I'm currently failing on Travis for 2 reasons. One is related to STYLE, I'm using clang-format v5 on my machine, would it make sense to update Travis' version of clang format from 4 to 5 as well?

For the second failure, I'll look at it closer once I fix the style issue.

AndrewSB:
I can build it 😄

fjricci:
Yeah, I mean, that's what I do. (I even have it as a command in my bashrc, which I find helpful: https://github.com/fjricci/configfiles/blob/1be83f5e813612d382c3c47d734d68efcc3c5d58/bash/bashrc#L28-L37) If you're going to be doing more toolchain work, you'll end up needing llvm builds anyway.

AndrewSB:
@sas I'll try to clean up my commits & their messages today

fjricci:
What's the status of this?

sas:
Bump.

AndrewSB:
This flew under my radar. Checking this out now! 

### TODO
- [ ] make sure the git commits are sensical
- [ ] rebase
- [ ] fix STYLE
- [ ] fix actual test breakages (like in https://travis-ci.org/facebook/ds2/jobs/269802643)

fjricci:
😮  TIL about TODO lists on github 😮

lanza:
no say bone

AndrewSB:
😱😥

PR #511, opened 2017-08-17
Implement software single-stepping on aarch64

fjricci:
Hardware single-stepping (slash ptrace stepping) is unsupported.

### DIFF
--- a/Headers/DebugServer2/Architecture/ARM/Branching.h
+++ b/Headers/DebugServer2/Architecture/ARM/Branching.h
@@ -101,8 +101,8 @@ enum BranchCond {
   kCondLS,
   kCondGE,
   kCondLT,
-  kCondLE,
   kCondGT,
+  kCondLE,
   kCondAL,
   kCondNV
 };
--- a/Sources/Architecture/ARM64/ARM64BranchInfo.cpp
+++ b/Sources/Architecture/ARM64/ARM64BranchInfo.cpp
@@ -1,0 +1,129 @@
+//
+// Copyright (c) 2014-present, Facebook, Inc.
+// All rights reserved.
+//
+// This source code is licensed under the University of Illinois/NCSA Open
+// Source License found in the LICENSE file in the root directory of this
+// source tree. An additional grant of patent rights can be found in the
+// PATENTS file in the same directory.
+//
+
+#include "DebugServer2/Architecture/ARM64/Branching.h"
+
+#include <cstdlib>
+
+using namespace ds2::Architecture::ARM64;
+
+namespace {
+
+class ARM64Instruction {
+  uint32_t _insn;
+
+public:
+  ARM64Instruction(uint32_t insn = 0) : _insn(insn) {}
+
+  void reset(uint32_t insn = 0) { _insn = insn; }
+
+  static inline int64_t SignExt(uint64_t n, size_t size) {
+    size_t const sh = sizeof(n) * CHAR_BIT - size;
+    return static_cast<int64_t>(n << sh) >> sh;
+  }
+
+  //
+  // B <imm>
+  // BL <imm>
+  // B.<cond> <imm>
+  //
+  inline bool getB(BranchInfo &info) const {
+    if ((_insn & 0x3c000000) == 0x14000000) {
+      uint64_t disp;
+      size_t size;
+      if ((_insn >> 30) & 1) {
+        info.type = kBranchTypeBcc;
+        info.cond = static_cast<BranchCond>(_insn & 0xf);
+        disp = (_insn & 0x00ffffe0) >> 5;
+        size = 21;
+      } else {
+        if ((_insn >> 31) & 1) {
+          info.type = kBranchTypeBL;
+        } else {
+          info.type = kBranchTypeB;
+        }
+        disp = _insn & 0x03ffffff;
+        size = 28;
+      }
+      info.disp = SignExt(disp << 2, size);
+      return true;
+    }
+    return false;
+  }
+
+  //
+  // BR <reg>
+  // BLR <reg>
+  // RET <reg>
+  //
+  inline bool getBR(BranchInfo &info) const {
+    if ((_insn & 0xfe1ffe1f) == 0xd61f0000) {
+      if ((_insn >> 22) & 0xf) {
+        info.type = kBranchTypeRET;
+      } else if ((_insn >> 21) & 1) {
+        info.type = kBranchTypeBLR;
+      } else {
+        info.type = kBranchTypeBR;
+      }
+      info.reg = (_insn & 0x0ff0) >> 5;
+
+      return true;
+    }
+    return false;
+  }
+
+  //
+  // CB{Z|NZ} <reg>, <imm>
+  // TB{Z|NZ} <reg>, #offset, <imm>
+  //
+  inline bool getBZ(BranchInfo &info) const {
+    if ((_insn & 0x7c000000) == 0x34000000) {
+      info.reg = _insn & 0x1f;
+      info.halfReg = !((_insn >> 31) & 1);
+
+      if ((_insn >> 25) & 1) {
+        info.type = kBranchTypeTB;
+        info.disp = SignExt(((_insn & 0x007fffe0) >> 5) << 2, 16);
+        info.offset = (_insn & 0x00f80000) >> 19;
+        if (!info.halfReg) {
+          info.offset += 32;
+        }
+      } else {
+        info.type = kBranchTypeCB;
+        info.disp = SignExt(((_insn & 0x00ffffe0) >> 5) << 2, 21);
+      }
+
+      return true;
+    }
+    return false;
+  }
+
+  bool getBranchInfo(BranchInfo &info) const {
+    info.cond = kCondNV;
+    info.halfReg = false;
+    info.reg = -1;
+    info.disp = 0;
+    info.offset = 0;
+
+    return (getB(info) || getBR(info) || getBZ(info));
+  }
+};
+} // namespace
+
+namespace ds2 {
+namespace Architecture {
+namespace ARM64 {
+
+bool GetARM64BranchInfo(uint32_t insn, BranchInfo &info) {
+  return ARM64Instruction(insn).getBranchInfo(info);
+}
+} // namespace ARM64
+} // namespace Architecture
+} // namespace ds2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,7 @@ set(ARCHITECTURE_ARM_SOURCES
 
 set(ARCHITECTURE_ARM64_SOURCES
     ${ARCHITECTURE_ARM_SOURCES}
+    Sources/Architecture/ARM64/SoftwareSingleStep.cpp
     Sources/Architecture/ARM64/RegistersDescriptors.cpp
     )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,7 @@ set(ARCHITECTURE_ARM_SOURCES
 
 set(ARCHITECTURE_ARM64_SOURCES
     ${ARCHITECTURE_ARM_SOURCES}
+    Sources/Architecture/ARM64/ARM64BranchInfo.cpp
     Sources/Architecture/ARM64/SoftwareSingleStep.cpp
     Sources/Architecture/ARM64/RegistersDescriptors.cpp
     )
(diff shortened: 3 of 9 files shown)


fjricci:
This works for step-in, but not step-over. I'm almost positive the step-over issues are lldb-side though.

xiaobai:
I cherry picked this change and tried it on my Linux-AArch64 board last night. Unfortunately both `step` and `next` commands didn't work (the entire program ran to completion).

fjricci:
Were you able to debug at all? Hard for me to repro without having a linux board

xiaobai:
I didn't look too much into it, I did it before I went to sleep. I can debug it later this evening when I get back to my apartment.

xiaobai:
tl;dr: getBranchInfo() in ARM64BranchInfo.cpp should reorder line 116 to be `return (getBR(info) || getB(info) || getBZ(info))` (and document why) or you should rewrite the get{B, BR, BZ})() functions so that they can't misidentify instructions based on the order they're checked in. We should probably go with the latter. I applied the reorder to confirm it would work, and I confirmed that step-over also fails for me. I also confirmed `finish` (step-out) fails with `error: Couldn't find thread plan to implement step type.`, so that's probably also on lldb's end.

I wrote a test that looks like this:
```cpp
int foo(int a) {
  int b = a * 2;
  return b + 1;
}

int main(int argc, char *argv[])
{
  int a = foo(1);
  int b = a * 2;
  int c = b - 1;
  return c;
}
```
I built this on a PINE64 running Ubuntu 16.04 with g++ ((Ubuntu/Linaro 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609)). Using objdump, I found the 2 functions assembled:
```
0000000000400790 <_Z3fooi>:
  400790:       d10083ff        sub     sp, sp, #0x20
  400794:       b9000fe0        str     w0, [sp,#12]
  400798:       b9400fe0        ldr     w0, [sp,#12]
  40079c:       531f7800        lsl     w0, w0, #1
  4007a0:       b9001fe0        str     w0, [sp,#28]
  4007a4:       b9401fe0        ldr     w0, [sp,#28]
  4007a8:       11000400        add     w0, w0, #0x1
  4007ac:       910083ff        add     sp, sp, #0x20
  4007b0:       d65f03c0        ret

00000000004007b4 <main>:
  4007b4:       a9bd7bfd        stp     x29, x30, [sp,#-48]!
  4007b8:       910003fd        mov     x29, sp
  4007bc:       b9001fa0        str     w0, [x29,#28]
  4007c0:       f9000ba1        str     x1, [x29,#16]
  4007c4:       52800020        mov     w0, #0x1                        // #1
  4007c8:       97fffff2        bl      400790 <_Z3fooi>
  4007cc:       b90027a0        str     w0, [x29,#36]
  4007d0:       b94027a0        ldr     w0, [x29,#36]
  4007d4:       531f7800        lsl     w0, w0, #1
  4007d8:       b9002ba0        str     w0, [x29,#40]
  4007dc:       b9402ba0        ldr     w0, [x29,#40]
  4007e0:       51000400        sub     w0, w0, #0x1
  4007e4:       b9002fa0        str     w0, [x29,#44]
  4007e8:       b9402fa0        ldr     w0, [x29,#44]
  4007ec:       a8c37bfd        ldp     x29, x30, [sp],#48
  4007f0:       d65f03c0        ret
```
ds2 is dying at address 0x4007b0, the `ret` instruction. I added some logs in PrepareARM64SoftwareSingleStep and it looks like the `ret` instruction is getting mis-identified as kBranchTypeBcc instead of kBranchTypeRET. In fact, the `br`, `blr`, and `ret` instructions will always get identified as a `b`, `bl`, or `b.c` instruction because the condition `_instr & 0x34000000 == 0x14000000` will always hold true. See for yourself:
```
Upper byte of instructions*:
b   -- 0001 01xx
bl  -- 1001 01xx
b.c -- 0101 0100
br  -- 1101 0110
blr -- 1101 0110
ret -- 1101 0110
```
This means we should call getBR() before getB(), or we should be more thorough when identify instructions. I think we should go with the latter, because which instruction gets identified shouldn't depend on the order in which you check them.

`*` Found [here](http://www.realworldtech.com/forum/?threadid=124953) and verified with [this](http://armconverter.com/).

xiaobai:
As for stepping-over not working properly, it looks like we set a breakpoint both at the next PC and the branch PC. Using the example from the previous post, I get this from ds2 logs:
```
[19729][PrepareSoftwareSingleStep] DEBUG  : PC=0x4007c8 branchPC=0x400790 [size=4, link=true] nextPC=0x4007cc [size=4]
[19729][SoftwareBreakpointManager::enableLocation] DEBUG  : set breakpoint instruction 0x002020d4 at 0x0000000000400790 (saved insn 0xff8300d1)
[19729][SoftwareBreakpointManager::enableLocation] DEBUG  : set breakpoint instruction 0x002020d4 at 0x00000000004007cc (saved insn 0xa02700b9)
```
On lldb's side, it looks like we hit the breakpoint set at 0x400790 and LLDB thinks we've finished stepping over. At least, I think that's what this says:
```
ThreadList::ShouldStop: 1 threads, 1 unsuspended threads
Thread::ShouldStop(0x653300) for tid = 0x4d12 0x4d12, pc = 0x0000000000400790
^^^^^^^^ Thread::ShouldStop Begin ^^^^^^^^
Plan stack initial state:
  thread #1: tid = 0x4d12:
    Active plan stack:
      Element 0: Base thread plan.
      Element 1: Stepping over line test.cpp:11 using ranges: [0x00000000004007c4-0x00000000004007d0).

Plan Step range stepping over explains stop, auto-continue 0.
ThreadPlanStepOverRange reached 0x0000000000400790.
ShouldStopHere callback returned 1 from 0x400790.
Plan Step range stepping over should stop: 1.
Completed step through range plan.
Popping plan: "Step range stepping over", tid = 0x4d12.
Plan stack final state:
  thread #1: tid = 0x4d12:
    Active plan stack:
      Element 0: Base thread plan.
    Completed Plan Stack:
      Element 0: Stepping over line test.cpp:11 using ranges: [0x00000000004007c4-0x00000000004007d0).

vvvvvvvv Thread::ShouldStop End (returning 1) vvvvvvvv
ThreadList::ShouldStop overall should_stop = 1
ThreadPlanStepOverRange reached 0x0000000000400790.
ShouldStopHere callback returned 1 from 0x400790.
Process 19730 stopped
```

0x400790 is the entry point of `foo()` and we don't really want to stop there if we're stepping over the call to `foo()`. So we shouldn't be putting a breakpoint there then, right?

fjricci:
I'm fairly certain that the stepping over issue is due to failed frame symbolication (the double breakpoint etc is the same as what we do on ARM). But good catch on the opcode parsing

fjricci:
Ooo those opcodes are useful

fjricci:
Although I'm impressed that I mostly reverse-engineered them correctly :D

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
```diff
@@ -0,0 +1,130 @@
...
+class ARM64Instruction {
+  uint32_t _insn;
+
+public:
+  ARM64Instruction(uint32_t insn = 0) : _insn(insn) {}
+
+  void reset(uint32_t insn = 0) { _insn = insn; }
+
+  static inline int64_t SignExt(uint64_t n, size_t size) {
+    size_t const sh = 64 - size;
```
compnerd: Would be nicer to do `sizeof(n) * CHAR_BIT` IMO, but meh.

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return false;
+  }
+
+  //
+  // BR <reg>
+  // BLR <reg>
+  // RET <reg>
+  //
+  inline bool getBR(BranchInfo &info) const {
+    if ((_insn & 0xff9ffc1f) == 0xd61f0000) {
```
compnerd: I think this should be `(_insn & 0xfe1ffe1f) == 0xd61f0000`.  That encodes the following instructions:

- blr (0x1)
- br (0x0)
- dret (0x5)
- eret (0x4)
- ret (0x2)
- ret

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
```diff
@@ -0,0 +1,130 @@
...
+    return static_cast<int64_t>(n << sh) >> sh;
+  }
+
+  //
+  // B <imm>
+  // BL <imm>
+  // B.<cond> <imm>
+  //
+  inline bool getB(BranchInfo &info) const {
+    if ((_insn & 0x7c000000) == 0x14000000) {
```
compnerd: I don't understand how this catches the `b.<cond>` case:

- b <imm> [0b000101TTTTTTTTTTTTTTTTTTTTTTTTTT]
- bl <imm> [0b100101TTTTTTTTTTTTTTTTTTTTTTTTTT]
- b.<cond> <imm> [0b01010100TTTTTTTTTTTTTTTTTTT0CCCC]

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  //
+  // BR <reg>
+  // BLR <reg>
+  // RET <reg>
+  //
+  inline bool getBR(BranchInfo &info) const {
+    if ((_insn & 0xff9ffc1f) == 0xd61f0000) {
+      if ((_insn >> 22) & 1) {
```
compnerd: The type is `(_insn >> 22) & 0xf` and the switch above.  Anything else is undefined.

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
```diff
@@ -0,0 +1,130 @@
...
+  //
+  // CB{Z|NZ} <reg>, <imm>
+  // TB{Z|NZ} <reg>, #offset, <imm>
+  //
+  inline bool getBZ(BranchInfo &info) const {
+    if ((_insn & 0x7c000000) == 0x34000000) {
+      info.reg = _insn & 0x1f;
+      info.halfReg = !((_insn >> 31) & 1);
+
+      uint16_t type = (_insn >> 24) & 0x0f;
```
compnerd: Why is `type` a `uint16_t`?.

type is `((_insn >> 25) & 1)` with 0 being `cbz` and 1 being `tbz`.

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
xiaobai: I'm pretty sure this still doesn't catch b.cond.

Headers/DebugServer2/Architecture/ARM/Branching.h
```diff
@@ -89,22 +89,22 @@ enum BranchDisp {
...
+  kCondVC = 0b0111,
+  kCondHI = 0b1000,
+  kCondLS = 0b1001,
+  kCondGE = 0b1010,
+  kCondLT = 0b1011,
+  kCondGT = 0b1100,
+  kCondLE = 0b1101,
+  kCondAL = 0b1110,
+  kCondNV = 0b1111
 };
```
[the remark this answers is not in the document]
fjricci: Well, I know they aren't necessary from a functionality perspective, but I also suspect that the original author wouldn't have made the mistake they made if they had explicitly labeled the values, since the explicit values make it easier to check for correctness. On the other hand, we now know they're correct, so I guess it's fine to remove them.

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
fjricci: You're right, it doesn't. Oops.

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
fjricci: The manual for aarch64 doesn't specify dret/eret as instructions that modify the pc. But perhaps I was looking at a manual for an older version of arm.

Sources/Architecture/ARM64/ARM64BranchInfo.cpp
fjricci: I figured uint16_t was a sufficiently small type. I could use a uint8_t though.

fjricci:
Thanks for the review @compnerd - you know too much 😆

fjricci:
Looks like this was a bug in android O

PR #515, opened 2017-08-22
[WIP] Enable tsan on travis checkin builds

fjricci:


sas:
I think we can close this for now. These scripts you're modifying are going away.

PR #516, opened 2017-08-23
Fix definition of the x87 ftag register

xiaobai:
This was recently fixed in LLDB (by @compnerd) in https://reviews.llvm.org/rL311579.

This should fix things in ds2.


xiaobai:
I don't think this change can go in until we switch over to LLDB 5.0, as this will break a decent amount of LLDB 4.0 tests.

fjricci:
Will it break just the register-reading tests or will it also break other behavior?

xiaobai:
I've been using this patch locally and it hasn't broken any behavior for me yet. It will probably just break the register reading tests.

fjricci:
Depending how many fail, you could add patch files.

sas:
Let's close the loop on this one. Should we add blacklists for register tests or should we just switch to using 5.0 for testing?

xiaobai:
I think we should switch to 5.0 for testing. The branch cut was ~1 month ago and it seems rather stable.

PR #518, opened 2017-09-12
Android AArch64/ARM64 support

xiaobai:
The long awaited automated testing of android aarhc64 support.

A few notes: an initial update of `platform-tools` and `tools` is not enough, for some reason. I think it's bootstrapped, so we should update them a second time to get the latest and greatest.


xiaobai:
Abandon all hope, ye who enter here.

PR #523, opened 2017-09-28

fjricci:

### DIFF
--- a/Sources/Core/BreakpointManager.cpp
+++ b/Sources/Core/BreakpointManager.cpp
@@ -45,11 +45,7 @@ ErrorCode BreakpointManager::add(Address const &address, Type type, size_t size,
     site.mode = mode;
     site.size = size;
 
-    // If the breakpoint manager is already in enabled state, enable
-    // the newly added breakpoint too.
-    if (enabled()) {
-      return enableLocation(site);
-    }
+    return enableLocation(site);
   }
 
   return kSuccess;
@@ -86,14 +82,7 @@ ErrorCode BreakpointManager::remove(Address const &address) {
 
 do_remove:
   DS2ASSERT(it->second.refs == 0);
-  //
-  // If the breakpoint manager is already in enabled state, disable
-  // the newly removed breakpoint too.
-  //
-  ErrorCode error = kSuccess;
-  if (enabled()) {
-    error = disableLocation(it->second);
-  }
+  ErrorCode error = disableLocation(it->second);
 
   _sites.erase(it);
   return error;
--- a/Sources/Core/BreakpointManager.cpp
+++ b/Sources/Core/BreakpointManager.cpp
@@ -113,43 +113,6 @@ void BreakpointManager::enumerate(
   }
 }
 
-void BreakpointManager::enable(Target::Thread *thread) {
-  //
-  // Both callbacks should be installed by child class before
-  // enabling the breakpoint manager.
-  //
-  if (enabled(thread)) {
-    DS2LOG(Warning, "double-enabling breakpoints");
-  }
-
-  enumerate([this, thread](Site const &site) { enableLocation(site, thread); });
-}
-
-void BreakpointManager::disable(Target::Thread *thread) {
-  if (!enabled(thread)) {
-    DS2LOG(Warning, "double-disabling breakpoints");
-  }
-
-  enumerate(
-      [this, thread](Site const &site) { disableLocation(site, thread); });
-
-  //
-  // Remove temporary breakpoints.
-  //
-  auto it = _sites.begin();
-  while (it != _sites.end()) {
-    it->second.type =
-        static_cast<Type>(it->second.type & ~kTypeTemporaryOneShot);
-    if (!it->second.type) {
-      // refs should always be 0 unless we have a kTypePermanent breakpoint.
-      DS2ASSERT(it->second.refs == 0);
-      _sites.erase(it++);
-    } else {
-      it++;
-    }
-  }
-}
-
 bool BreakpointManager::hit(Address const &address, Site &site) {
   if (!address.valid())
     return false;
--- a/Sources/Core/HardwareBreakpointManager.cpp
+++ b/Sources/Core/HardwareBreakpointManager.cpp
@@ -128,11 +128,6 @@ ErrorCode HardwareBreakpointManager::addThread(Target::Thread *thread) {
   return kSuccess;
 }
 
-// Hardware breakpoints are always enabled
-bool HardwareBreakpointManager::enabled(Target::Thread *thread) const {
-  return true;
-}
-
 bool HardwareBreakpointManager::fillStopInfo(Target::Thread *thread,
                                              StopInfo &stopInfo) {
   BreakpointManager::Site site;
--- a/Sources/Core/HardwareBreakpointManager.cpp
+++ b/Sources/Core/HardwareBreakpointManager.cpp
@@ -40,12 +40,13 @@ ErrorCode HardwareBreakpointManager::add(Address const &address, Type type,
 }
 
 ErrorCode HardwareBreakpointManager::remove(Address const &address) {
+  CHK(super::remove(address));
+
   auto loc = std::find(_locations.begin(), _locations.end(), address);
   if (loc != _locations.end()) {
     _locations[loc - _locations.begin()] = 0;
   }
-
-  return super::remove(address);
+  return kSuccess;
 }
 
 ErrorCode HardwareBreakpointManager::enableLocation(Site const &site,
@@ -64,9 +65,8 @@ ErrorCode HardwareBreakpointManager::enableLocation(Site const &site,
   }
 
   enumerateThreads(thread, [&](Target::Thread *t) {
-    if (t->state() == Target::Thread::kStopped && !enabled(t)) {
-      error = enableLocation(site, idx, t);
-    }
+    DS2ASSERT(t->state() == Target::Thread::kStopped);
+    error = enableLocation(site, idx, t);
   });
 
   if (error == kSuccess) {
@@ -86,9 +86,8 @@ ErrorCode HardwareBreakpointManager::disableLocation(Site const &site,
   int idx = loc - _locations.begin();
 
   enumerateThreads(thread, [&](Target::Thread *t) {
-    if (t->state() == Target::Thread::kStopped && enabled(t)) {
-      error = disableLocation(idx, t);
-    }
+    DS2ASSERT(t->state() == Target::Thread::kStopped);
+    error = disableLocation(idx, t);
   });
 
   return error;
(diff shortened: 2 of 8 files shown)


fjricci:
Verified this works:

`
(lldb) b 0x4004fc
Breakpoint 2: where = test`main + 15 at test.cc:4, address = 0x00000000004004fc
(lldb) disass --bytes
test`main:
    0x4004ed <+0>:  55                    pushq  %rbp
    0x4004ee <+1>:  48 89 e5              movq   %rsp, %rbp
->  0x4004f1 <+4>:  c7 45 fc 00 00 00 00  movl   $0x0, -0x4(%rbp)
    0x4004f8 <+11>: 83 45 fc 01           addl   $0x1, -0x4(%rbp)
    0x4004fc <+15>: 8b 45 fc              movl   -0x4(%rbp), %eax
    0x4004ff <+18>: 83 e8 01              subl   $0x1, %eax
    0x400502 <+21>: 5d                    popq   %rbp
    0x400503 <+22>: c3                    retq
(lldb) x 0x4004fc
0x004004fc: 8b 45 fc 83 e8 01 5d c3 66 2e 0f 1f 84 00 00 00  .E....].f.......
0x0040050c: 00 00 66 90 41 57 41 89 ff 41 56 49 89 f6 41 55  ..f.AWA..AVI..AU
(lldb)
`

fjricci:
```
(lldb) b 0x4004fc
Breakpoint 2: where = test`main + 15 at test.cc:4, address = 0x00000000004004fc
(lldb) disass --bytes
test`main:
    0x4004ed <+0>:  55                    pushq  %rbp
    0x4004ee <+1>:  48 89 e5              movq   %rsp, %rbp
->  0x4004f1 <+4>:  c7 45 fc 00 00 00 00  movl   $0x0, -0x4(%rbp)
    0x4004f8 <+11>: 83 45 fc 01           addl   $0x1, -0x4(%rbp)
    0x4004fc <+15>: 8b 45 fc              movl   -0x4(%rbp), %eax
    0x4004ff <+18>: 83 e8 01              subl   $0x1, %eax
    0x400502 <+21>: 5d                    popq   %rbp
    0x400503 <+22>: c3                    retq
(lldb) x 0x4004fc
0x004004fc: 8b 45 fc 83 e8 01 5d c3 66 2e 0f 1f 84 00 00 00  .E....].f.......
0x0040050c: 00 00 66 90 41 57 41 89 ff 41 56 49 89 f6 41 55  ..f.AWA..AVI..AU
(lldb)
```

Sources/Core/SoftwareBreakpointManager.cpp
```diff
@@ -94,23 +94,34 @@ ErrorCode SoftwareBreakpointManager::disableLocation(Site const &site,
...
+  while (it != _sites.end()) {
+    it->second.type =
+        static_cast<Type>(it->second.type & ~kTypeTemporaryOneShot);
+    if (!it->second.type) {
+      // refs should always be 0 unless we have a kTypePermanent breakpoint.
+      DS2ASSERT(it->second.refs == 0);
+      _sites.erase(it++);
+    } else {
+      it++;
+    }
```
xiaobai: You could rewrite this to be something like:
```
if (!it->second.type) {
  // blah
  DS2ASSERT(bloop);
  _sites.erase(it);
}
it++;
```

This feels cleaner than having `it++` twice. Thoughts?

Sources/Target/Common/ProcessBase.cpp
```diff
@@ -225,6 +225,11 @@ ErrorCode ProcessBase::readMemoryBuffer(Address const &address, size_t length,
     return error;
   }
 
+  SoftwareBreakpointManager *bpm = softwareBreakpointManager();
+  if (bpm) {
+    bpm->insertStashedInsns(address, length, buffer);
+  }
+
```
xiaobai: If I understand correctly, you want to make sure that when we perform a memory read, we're not sending the breakpoint instructions we potentially placed. You want to return the actual instruction that is supposed to be there. Is this right?

Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -64,9 +65,8 @@ ErrorCode HardwareBreakpointManager::enableLocation(Site const &site,
   }
 
   enumerateThreads(thread, [&](Target::Thread *t) {
-    if (t->state() == Target::Thread::kStopped && !enabled(t)) {
-      error = enableLocation(site, idx, t);
-    }
+    DS2ASSERT(t->state() == Target::Thread::kStopped);
```
xiaobai: Why did this become an assert? I see you also do it in disableLocations.
Are all threads supposed to be stopped when enabling a new hardware breakpoint?

Sources/Core/SoftwareBreakpointManager.cpp
fjricci: Yep, I meant to change that and forgot, good catch.

Sources/Target/Common/ProcessBase.cpp
fjricci: Yep

Sources/Core/HardwareBreakpointManager.cpp
fjricci: We discussed it a bit with stephane yesterday on messenger, but basically this function is now only run on breakpoint creation or thread creation, both of which assume the threads are stopped.

fjricci: Or at least, the thread in question.

Sources/Core/SoftwareBreakpointManager.cpp
sas: Don't. After `_sites.erase(it)`, the iterator becomes invalid, so you can't do `it++`.

fjricci: 😮

xiaobai: Shows how much I know! :D

fjricci:
Judging my my local VM, I think it's TestLongJump:test_step_over that's hanging. It's an XFAIL anyway, so I'll move it to a SKIP.

fjricci:
Hanging test is StepOverBreakpoint

fjricci:
Repro with:
LINUX_DISTRO=trusty TARGET=Android-ARM LLDB_TESTS=StepOverBreak PLATFORM=1 ./Support/Testing/Travis/script.sh

Sources/Core/BreakpointManager.cpp
```diff
@@ -90,10 +86,7 @@ ErrorCode BreakpointManager::remove(Address const &address) {
   // If the breakpoint manager is already in enabled state, disable
   // the newly removed breakpoint too.
   //
```
xiaobai: Can you update/remove the comments here too? Doesn't seem accurate anymore

Sources/Core/SoftwareBreakpointManager.cpp
```diff
@@ -124,4 +124,31 @@ bool SoftwareBreakpointManager::fillStopInfo(Target::Thread *thread,
...
+
+    // overlap begins at beginning of instruction
+    if (start <= insn_start) {
+      uint64_t offset = insn_start - start;
+      uint64_t size = std::min((uint64_t)insn.second.size(), end - insn_start);
+      memcpy(&data[offset], &insn.second[0], size);
+    } else {
+      DS2ASSERT(start < insn_end);
+      uint64_t offset = insn_end - start;
+      uint64_t size = start - insn_end;
```
xiaobai: It should be the case that `start` is strictly less than `insn_end`, as shown by the assert and continue above. This result will be negative, which I'm sure is not what you wanted.

I think the correct thing to do here is `insn_end - start`, since that should be the number of bytes you want to write in this case.

Sources/Core/SoftwareBreakpointManager.cpp
```diff
@@ -124,4 +124,31 @@ bool SoftwareBreakpointManager::fillStopInfo(Target::Thread *thread,
...
+    }
+
+    // overlap begins at beginning of instruction
+    if (start <= insn_start) {
+      uint64_t offset = insn_start - start;
+      uint64_t size = std::min((uint64_t)insn.second.size(), end - insn_start);
+      memcpy(&data[offset], &insn.second[0], size);
+    } else {
+      DS2ASSERT(start < insn_end);
+      uint64_t offset = insn_end - start;
```
xiaobai: I'm pretty sure this should be `start - insn_start`, since you want the index into the instruction to start writing?

For example, I think it will fail here:
```
insn_start := 0
start := 1
insn_end := 4
```
The instruction is 4 bytes long. Your code calculates the offset to be 4 -  1 = 3, which means you only write the last byte in the instruction when you wanted to copy the last 3 bytes.

I might be wrong, so you should check my understanding of this.

Sources/Target/Common/ProcessBase.cpp
```diff
@@ -304,21 +301,16 @@ ErrorCode ProcessBase::afterResume() {
     return kSuccess;
   }
 
-  // Disable breakpoints and try to hit software breakpoints.
-  BreakpointManager *swBpm = softwareBreakpointManager();
-  if (swBpm != nullptr) {
+  // Try to hit software breakpoints and clear temporary breakpoints.
+  SoftwareBreakpointManager *bpm = softwareBreakpointManager();
+  if (bpm != nullptr) {
```
xiaobai: nit: You can change this into `if (bpm)`

Sources/Target/Common/ThreadBase.cpp
```diff
@@ -32,9 +32,9 @@ ErrorCode ThreadBase::modifyRegisters(
 }
 
 ErrorCode ThreadBase::beforeResume() {
-  BreakpointManager *bpm = _process->hardwareBreakpointManager();
+  HardwareBreakpointManager *bpm = _process->hardwareBreakpointManager();
   if (bpm != nullptr) {
```
xiaobai: I know this isn't your code, but you could change it to `if (bpm` as well here.

Sources/Core/HardwareBreakpointManager.cpp
```diff
@@ -40,12 +40,13 @@ ErrorCode HardwareBreakpointManager::add(Address const &address, Type type,
 }
 
 ErrorCode HardwareBreakpointManager::remove(Address const &address) {
+  CHK(super::remove(address));
```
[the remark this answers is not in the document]
fjricci: Yeah, I believe that's the case (ugh I wrote this patch 8 months ago T_T).

Sources/Core/SoftwareBreakpointManager.cpp
```diff
@@ -124,4 +124,30 @@ bool SoftwareBreakpointManager::fillStopInfo(Target::Thread *thread,
...
+      uint64_t offset = insn_start - start;
+      uint64_t size = std::min((uint64_t)insn.second.size(), end - insn_start);
+      memcpy(&data[offset], &insn.second[0], size);
+    } else {
+      uint64_t offset = start - insn_start;
+      uint64_t size = insn_end - start;
+      memcpy(&data[0], &insn.second[offset], size);
+    }
+  }
+}
```
xiaobai: Something I just thought about: Is it possible for and instruction size to be greater than `end - start`? By that, I mean is this possible: `insn_start <= start <= end <= insn_end`?

My gut says it's possible but unlikely. If so, the `size` in the else path will be larger than you're allowed to write and you might segfault. It should be something like `std::min(insn_end - start, end - start)`, I think.

sas:
Moved to PR #641.

PR #524, opened 2017-09-29
Fix register clobbering in Linux-aarch64 mmap code preparation

xiaobai:
This was preventing us from allocating memory in the remote process being debugged occasionally. Sometimes we would luck out, allocate memory, and segfault instead :D. This should prevent that the allocation bug! :D


fjricci:
🙃🙃

0xsaipal:
lol

PR #526, opened 2017-10-06
Replacing gcc with ehframe for qRegisterInfo packets

krishnamittal96:
Basically, lldb just changed the name of the string in their communication protocol, and "gcc" needs to be replaced with "ehframe". This commit also changes name  of variable used to hold the gcc register index value, since now we use it to store ehframe register index value.
See Task #T10727553 for more information.


fjricci:
As a followup, we should probably trim the deprecated "gcc"-style values from lldb, since they were only in there for us I think. but probably want to wait until after the 6.0 release (so @krishnamittal96 doesn't need to worry, just documenting here)

fjricci:
Please squash the commits (`git rebase -i`)

krishnamittal96:
Also will squash all commits once I have fixed the style error.

fjricci:
You could also try running script.sh with TARGET=Style locally

krishnamittal96:
I was finally able to test locally. I had to use ninja instead of make and install ncurses and swig. I am getting no style errors now.

fjricci:
I think you should actually be able to move all of the gcc variable names to ehframe, I don't see any uses where that wouldn't be possible

krishnamittal96:
The registers test seems to be fine. The errors are shown in in autogenerated files. The includes in my file are in alphabetical order, but the test wants them not to be in alphabetical order. Also there are just some additional comments which were autogenerated.

I am not sure why one of the linux tests is failing. I double checked my code, and that seems to be the only thing I can do. I don't have access to osmeta, and can't seem to see the python test cases that fail

Tools/RegsGen2/RegisterTemplate.cpp
```diff
@@ -358,17 +358,17 @@ Register *RegisterTemplate::make(std::string const &name,
...
         fprintf(stderr, "error: register '%s' specifies a negative "
-                        "GCC register number",
+                        "EHFRAME register number",
                 name.c_str());
         return nullptr;
       }
 
-      _GCCRegisterNumber.mark(base->value());
-      reg->GCCRegisterNumber = base->value();
+      _EHFRAMERegisterNumber.mark(base->value());
```
fjricci: I think this should be _ehframeRegisterNumber

Tools/RegsGen2/RegisterTemplate.cpp
```diff
@@ -358,17 +358,17 @@ Register *RegisterTemplate::make(std::string const &name,
     reg->GDBRegisterNumber = base->value();
   }
 
-  if (!_DWARFGCCAliased) {
-    if (auto base = d->value<JSInteger>("gcc-reg-number")) {
+  if (!_DWARFEHFRAMEAliased) {
```
fjricci: _DWARFEhframeAliased

fjricci:
I'm also going to be on PTO the rest of the week, so @sas  will probably need to take over the review

krishnamittal96:
clang-format-4.0: command not found
I have clang-format version 6.0 and travis seems to be running clang-format-4.0

krishnamittal96:
All checks pass, and all changes requested by Francis were made. Waiting for review @sas

PR #527, opened 2017-10-09
More work with platform mode initialization

sas:


Sources/main.cpp
```diff
@@ -83,6 +84,83 @@ CreateSocket(std::string const &host, std::string const &port, bool reverse) {
   return socket;
 }
 
+#if defined(OS_POSIX)
+static std::unique_ptr<Socket> CreateUNIXSocket(std::string const &path,
+                                                bool abstract, bool reverse) {
+#if !defined(OS_LINUX)
+  if (abstract) {
+    DS2LOG(Fatal, "abstract UNIX sockets not supported on this system");
```
xiaobai: Isn't this already caught in `Socket::Listen`? Is there a need to catch this in both places?

PR #531, opened 2017-10-12
Change Travis config for Android

xiaobai:
We should re-enable android-x86 builds for API level 2x (so we're not just testing if it builds for API level 16), and we can get testing for android-arm64. Woohoo!


xiaobai:
We're moving to CircleCI. After we migrate, we can re-evaluate.

PR #532, opened 2017-10-13
Fix mmap protections for POSIX systems

sas:


Headers/DebugServer2/Target/POSIX/Process.h
```diff
@@ -40,6 +40,10 @@ class Process : public ds2::Target::ProcessBase {
   ErrorCode writeMemory(Address const &address, void const *data, size_t length,
                         size_t *count = nullptr) override;
 
+protected:
+  int convertMemoryProtectionToPOSIX(uint32_t protection) const;
```
xiaobai: This should return a uint32_t. Most invocations of PrepareMmapCode accept a `uint32_t` for permissions, not an `int`. This shouldn't explicitly be a problem, but I'd prefer consistency in types here.

xiaobai:
Looks like you changed all the PrepareMmapCode functions to take an int instead of an int32_t. This is also fine, type consistency makes me happy :D

PR #534, opened 2017-10-17
Aarch64 mmap fixes

sas:


Headers/DebugServer2/Host/Linux/ARM64/Syscalls.h
```diff
@@ -34,6 +34,13 @@ static inline uint32_t MakeMovNegImmInstr(uint8_t reg, int32_t val) {
   return base | (static_cast<uint16_t>(-val - 1) << 5) | reg;
 }
 
+static inline uint32_t MakeLdrRelInstr(uint8_t reg, int32_t offset) {
+  DS2ASSERT(reg <= 31);
+  DS2ASSERT(offset % 4 == 0);
+  static const uint32_t base = 0x58000000;
+  return base | (static_cast<uint16_t>(offset / 4) << 5) | reg;
```
xiaobai: This is interesting. I think we could simply do `offset << 3`, but I like the preservation of intent. The compiler also is probably smart enough to figure out this optimization anyway.

sas:
The compiler is most definitely smart enough to figure this out, yes.
Actually, even if it wasn't, it wouldn't be a big deal here, but that's
another discussion :)

As you guessed, the goal here is to make the code more explicit.

> --
-- 
Stephane Sezer

PR #535, opened 2017-10-18
Add CircleCI

AndrewSB:
I think it would be a good idea to run both of these simultaneously for a little while to see if CircleCI is significantly faster


sas:
I think you'd have to enable circleCI for this repo for this PR to do anything diffirently. No?

AndrewSB:
T22783976

sas:
I'm gonna move this to PR #566.

PR #539, opened 2017-11-17
Adjust wrapper for RS3 headers

smeenai:
The RS3 headers expose far more APIs to store apps. Having duplicate
struct definitions causes compile errors. Having duplicate prototypes
doesn't cause compile errors, but it's still unclean and unnecessary.

Unfortunately, the Windows SDK headers don't expose any preprocessor
macro containing the SDK version number. We could use the presence of
certain other macros in the Windows headers (e.g. `NTDDI_WIN10_RS3`) as
a proxy for the SDK version, but that's ugly and significantly less
granular than the SDK version (since we'll only have one `NTDDI` macro
per Windows 10 major release, whereas there are many prerelease SDKs).
Instead, we can determine the SDK version from the environment and pass
it along to the preprocessor. This relies on the user invoking cmake
from a Visual Studio command prompt, which seems like a reasonable
assumption.

I didn't go through previous Windows SDK headers to figure out exactly
which version of the SDK these definitions were first exposed in, so the
check against 16299 might not be accurate in all cases, but it's good
enough for targeting the RS3 SDK. Other users who care about other SDK
versions can tweak the conditionals further as needed.

I also didn't bother consolidating definitions together to reduce the
amount of conditional blocks. That can be done in a follow-up if
desired.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,8 +318,17 @@ set_property(SOURCE Sources/main.cpp APPEND PROPERTY COMPILE_DEFINITIONS
 
 if ("${OS_NAME}" MATCHES "Windows")
   target_compile_definitions(ds2 PRIVATE WIN32_LEAN_AND_MEAN NOMINMAX CINTERFACE)
-  target_compile_definitions(ds2 PRIVATE WINVER=_WIN32_WINNT_WIN6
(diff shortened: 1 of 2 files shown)


sas:
I don't think this is the right approach to the problem. I'm fixing this in PR#542.

PR #541, opened 2017-11-20
Enable VS 2017 builds for Windows

sas:


xiaobai:
It looks like you can't just use "Visual Studio 2015" for the generator, it needs to be something like "Visual Studio 14 2015" instead. Same thing for VS2017. :(

.appveyor.yml
```diff
@@ -1,5 +1,7 @@
 # Operating system (build VM template)
-os: Visual Studio 2015
+image:
+  - Visual Studio 2015
```
xiaobai: `Visual Studio 14 2015`

.appveyor.yml
```diff
@@ -1,5 +1,7 @@
...
+  - Visual Studio 2017
```
xiaobai: `Visual Studio 15 2017`

PR #545, opened 2017-12-07
Diagnosing travis failure

xiaobai:
Travis is failing and we don't know why. Log all the things. Maybe we should implement a travis debugger?


xiaobai:
It looks like it was a transient infra failure. :/

PR #549, opened 2017-12-12
Do not pass-through SIGINT by default

sas:
This was passed through by default, I assume to help with interactive
apps where the user might want to quit the app while it is being
debugged, but it seems to create problems on Android because of app
lifecycle management.


sas:
This won't be required.

PR #552, opened 2017-12-14
Move build and test infra to LLVM 5.0

xiaobai:
We're still on 4.0, which is bad for us. Let's see what travis thinks? I'm sure there will be test failures, let's address them as they come.

I'm picking up the work initially done by AndrewSB in #525, with some modifications.


xiaobai:
Working on fixing Travis failures unrelated to moving our LLVM branch forward in #555. I'll get back to this right after.

fjricci:
Usually when we upgrade toolchains, at least some of the unit tests get fixed. Did you check the XPASS's from the travis results to see if we can remove anything from the blacklists?

xiaobai:
I did not do that because I intend on doing that after this PR. My reasoning is that some of the tests are blacklisted because they are flaky. I'd like to be first get us tested on 5.0 before I start re-enabling tests, so that we can take our time with re-enabling tests without being worried about moving forward.

Support/Testing/Travis/install.py
```diff
@@ -65,6 +65,7 @@
     android_platform = os.getenv('ANDROID_PLATFORM', "21")
     check_call('./Support/Scripts/prepare-android-toolchain.sh "%s:%s"' %
                (android_toolchains[target], android_platform), shell=True)
+    check_call('./Support/Scripts/prepare-android-ndk.sh', shell=True)
```
sas: This should be in the next commit instead, no?

xiaobai: It could be, but I added it in this commit to make sure the script actually worked. That is to say, to make sure that the script didn't error out and die at some point. I don't think it's a big deal which commit it goes into.

xiaobai:
Looks like coveralls doesn't like what we're doing. I won't merge until I fix that.

xiaobai:
I squashed some commits and re-ordered them in an attempt to make bisecting less painful in the future.

PR #553, opened 2017-12-15
Remove CI builds for gcc on Darwin

xiaobai:
I'm under the impression most people are going to be building things on Darwin with Clang, as its a part of the default toolchain on OSX. I don't expect gcc to fail to build ds2 on Darwin if Clang works. 

Additionally, the Mac queues on Travis are frequently backed up and are a bottleneck in CI signal. Minimizing the amount of work we need to do on OSX is better imo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,13 +79,7 @@ matrix:
     # Darwin-X86_64 targets
     - os: osx
       osx_image: xcode8.3
-      env: TARGET="Darwin-X86_64" CLANG="0"
-    - os: osx
-      osx_image: xcode8.3
       env: TARGET="Darwin-X86_64" CLANG="1"
-    - os: osx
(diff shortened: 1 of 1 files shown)


sas:
We do get some signal from these sometimes (gcc can refuse to compile code that clang accepts and vice-versa) but if it's too much of a pain for queue times, I'm fine with getting rid of these.

xiaobai:
Good to know. I do value the signal we'd get, but I value faster CI times more right now.

fjricci:
Would be nice if travis had a way to differentiate build types (ie weekly cronjob vs checkin build)

xiaobai:
It looks like we can differentiate if we check for the environment variable `TRAVIS_EVENT_TYPE`. However, I don't think that will stop the jobs from being spawned.

PR #554, opened 2017-12-20
Diagnose android failure on Travis

xiaobai:
Travis is failing when we play with android emulators. I want to use this PR to figure out why.


xiaobai:
I'm working on a solution I intend on committing in #555.

PR #555, opened 2018-01-05
Update scripts to fix Travis tests

xiaobai:
The adb used by travis is too old to work with the most up-to-date emulators in the android SDK. We should use the adb that ships with the SDK instead.

### DIFF
--- a/Support/Scripts/common.sh
+++ b/Support/Scripts/common.sh
@@ -69,3 +69,11 @@ linux_distribution() {
     echo "unknown"
   fi
 }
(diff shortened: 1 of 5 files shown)


Support/Scripts/common.sh
```diff
@@ -69,3 +69,11 @@ linux_distribution() {
     echo "unknown"
   fi
 }
+
+get_host_platform_name() {
```
fjricci: 😮  a function!

PR #556, opened 2018-01-09
Re-enable Android-X86 testing

sas:


xiaobai:
It looks like ds2 hit's a FATAL and crashes in the middle of running tests. Line 3433 of this Travis run: https://travis-ci.org/facebook/ds2/jobs/327024293

That being said, I'm sure we *can* turn this back on and get good feedback.

sas:
Yeah so this doesn't work because we don't have KVM on Travis. What a PITA...

xiaobai:
Ahh right, I forgot about that. I also linked to the Android-ARM run because I'm very attentive :D 

Anyway, there might be a way to disable hardware acceleration when running. Perhaps this is still relevant? https://stackoverflow.com/a/34054673

xiaobai:
Actually doesn't look like there's a feasible way to do this cleanly, if at all. :/

xiaobai:
Talked to @sas offline, he had an idea to try and run android-x86 tests on Mac machines instead. I think this should be fine, since the SDK probably ships with everything we need to run the x86 emulators on Darwin? It's worth a shot.

sas:
I'll take care of this along with the move to CircleCI.

PR #559, opened 2018-01-17
Re-enable some FP register tests

sas:
These were disabled because the behavior for some FP regs was broken in
lldb 4.0 and got fix for lldb 5.0. Now that we use 5.0, we can re-enable
these.


sas:
There are only two darwin builds that didn't finish yet. Everything else built and passed tests so I'm gonna land this.

PR #569, opened 2018-01-27
Move Linux builds to CircleCI

sas:


.circleci/config.yml
```diff
@@ -26,9 +26,16 @@ build_step: &build_step
 test_step: &test_step
   run:
     name: Test
-    command: test -z "$LLDB_TESTS" || (cd build && ../Support/Scripts/run-lldb-tests.sh)
+    command: cd build && ../Support/Scripts/run-lldb-tests.sh
```
xiaobai: why no more test?

xiaobai: Oh wait, because we actually want to perform the tests now?

sas:
Yeah. Mostly because it’s cleaner in that format.
-- 
-- 
Stephane Sezer

.circleci/config.yml
sas: Yeah we include the test step only when we actually want to run tests.

PR #572, opened 2018-01-27
Simplify ELFSupport::MachineTypeToCPUType

xiaobai:
I did this for two reason:
1) It felt strange to me that we were updating the value of `type` and `subType` if we resolved the CPUType correctly, but we otherwise left these alone... except 2 of the 3 calls changed the values to some `kCPUTypeAny` and `kCPUSubTypeInvalid` anyway. We can move these into `MachineTypeToCPUType` and avoid these 2 branches entirely.
2) Following (1), we no longer use the bool result except at one call site. It felt "dirty" to me that we were using the result of the call in some places and outright ignoring it elsewhere, so I changed the return type to `void` and instead examined what value the `type` reference had. I'm wondering what y'all think about this one.

Thoughts?

### DIFF
--- a/Sources/Host/Linux/ProcFS.cpp
+++ b/Sources/Host/Linux/ProcFS.cpp
@@ -635,11 +635,7 @@ CPUType ProcFS::GetProcessCPUType(pid_t pid) {
 
   CPUType type;
   CPUSubType subtype;
-  if (!ELFSupport::MachineTypeToCPUType(info.machine, info.is64Bit, type,
-                                        subtype)) {
-    type = kCPUTypeAny;
-  }
-
+  ELFSupport::MachineTypeToCPUType(info.machine, info.is64Bit, type, subtype);
   return type;
 }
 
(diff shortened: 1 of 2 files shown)


xiaobai:
sick, travis passed

Sources/Target/POSIX/ELFProcess.cpp
```diff
@@ -374,8 +374,9 @@ ErrorCode ELFProcess::updateInfo() {
   // we also validate that this ELF file is really for the current
   // architecture.
   //
-  if (!ELFSupport::MachineTypeToCPUType(machineType, is64Bit, _info.cpuType,
-                                        _info.cpuSubType))
+  ELFSupport::MachineTypeToCPUType(machineType, is64Bit, _info.cpuType,
+                                   _info.cpuSubType);
+  if (_info.cpuType == kCPUTypeAny)
```
xiaobai: Is it a problem to still populate _info's fields if ELFSupport::MachineTypeToCPUType fails?

sas: I think what you did here is good. Keeping the early return is probably what we want.

xiaobai:
@sas Can you re-review this when you get a chance?

PR #574, opened 2018-01-30
add --fd option

yuex:


Sources/main.cpp
```diff
@@ -260,6 +260,8 @@ static int GdbserverMain(int argc, char **argv) {
                  "connect back to the debugger at [HOST]:PORT");
   opts.addOption(ds2::OptParse::boolOption, "native-regs", 'r',
                  "use native registers (no-op)", true);
+  opts.addOption(ds2::OptParse::stringOption, "fd", 'F',
+                 "use a fild descriptor to communicate");
```
xiaobai: `fild` -> `file`

Sources/main.cpp
```diff
@@ -318,27 +320,35 @@ static int GdbserverMain(int argc, char **argv) {
...
+      socket = CreateTCPSocket(
+          gDefaultHost, namedPipePath.empty() ? gDefaultPort : "0", reverse);
     } else {
-      // Write the null terminator to the file. This follows the llgs behavior.
-      fwrite(realPort.c_str(), 1, realPort.length() + 1, namedPipe);
-      fclose(namedPipe);
+      socket = CreateSocket(opts.getPositional("[host]:port"), reverse);
+    }
+
+    if (!namedPipePath.empty()) {
```
xiaobai: If we pass the `--fd` argument, can we also use a named pipe? I'm kind of a noob when it comes to this part of ds2.

If we can combine these options, we can reformat the above code to be an `if, elif, else` block (creating a socket), and move the below block (using a named pipe) one scope out. Otherwise, this is fine.

Sources/main.cpp
```diff
@@ -355,8 +365,12 @@ static int GdbserverMain(int argc, char **argv) {
   else
     impl = ds2::make_unique<DebugSessionImpl>();
 
-  return RunDebugServer(reverse ? socket.get() : socket->accept().get(),
-                        impl.get());
+  if (fd >= 0) {
+    return RunDebugServer(socket.get(), impl.get());
```
xiaobai: It'd be nice if we could avoid having 2 return statements. As in, `RunDEbugServer` takes 2 argument and the second one is the same each time. It'd be nice if we could use this if/else to choose just the first argument and call `RunDebugServer` under it.

But do we actually need to have this? I'm not sure because I don't have some knowledge on the subject, but is it possible to combine reverse connecting and socket pairs?

yuex:
@xiaobai thanks for reviewing my code. I test it buy setting up `LLDB_DEBUGSERVER_PATH` to ds2 and fire lldb to see if it can run a simple binary executable. But it didn't pass the test. I talked with @sas. We think it's a good idea to initiate a PR so we can double check the code online.

Sources/main.cpp
yuex: I think `--fd` will override `--named-pipe`. because we only use the named pipe to write the dynamically allocated port. but when fd is used, we don't even have a port. so I think fd should override host:port as well as named pipe

Sources/main.cpp
yuex: i don't think so. reverse controls how the sockfd is created, by accept() or by connect(). but when fd is used, we already have the fd. it doesn't make sense to allow reverse anymore.

definitely, i can combine the two branches into one statement using `?:`

Sources/main.cpp
sas: @yuex is right, except I don't think `--fd` should necessarily override `--named-pipe`, but we probably want to display an error message instead.

yuex: cool, I can add this

yuex:
yes, it still fails after removing `close(i)`.

another thing I observed is that if I run `ds2 g --fd 1`, after hit enter once, ds2 will exit. when running `ds2 g --fd 100` where fd 100 doesn't exist, ds2 will exit directly. Is this behavior expected?

sas:
@yuex: I don't think that's a good behavior. What we should do instead is check the file descriptor we are passed in. You can use `ioctl()` to check if the file descriptor is valid. Another check you might wanna do is call `isatty()` on the file descriptor to make sure the user isn't passing down 0, 1, or 2 with `--fd`.

yuex:
it seems that MinGW doesn't have fd at all and doesn't support MSG_DONTWAIT flag in recv(). Right now, I just made fd option have no effect when OS_POSIX is not set. Another way is to print an error message when fd option is used on windows. Do you think this is necessary? Always better to fail loudly?

yuex:
on Linux, 3 leak tests failed. Looks like we need to keep the close(i) logic when fd option is not used.

Sources/main.cpp
```diff
@@ -490,16 +557,7 @@ static int VersionMain(int argc, char **argv) {
 __declspec(dllexport)
 #endif
 int main(int argc, char **argv) {
-// clang-format on
-
-#if defined(OS_POSIX)
```
sas: I think you should just move this block below, after the argument parsing is done, and add a simple `if ()` statement in there to avoid closing the file descriptor you get passed by the `--fd` argument.

Sources/main.cpp
```diff
@@ -511,9 +569,17 @@ int main(int argc, char **argv) {
     UsageDie(argv[0]);
   }
 
-  switch (argv[1][0]) {
-  case 'g':
+  if (argv[1][0] == 'g')
```
sas: You can move the close loop before this line and leave the `switch ()` block untouched.

Sources/main.cpp
```diff
@@ -50,6 +50,40 @@ static std::string gDefaultHost = "127.0.0.1";
 static bool gDaemonize = false;
 static bool gGDBCompat = false;
 
+#if defined(OS_POSIX)
+static void CloseFD() {
```
sas: These two functions seem unnecessary, you can just get rid of them and move the close-loop further down in the main() function.

Sources/main.cpp
```diff
@@ -50,6 +50,40 @@ static std::string gDefaultHost = "127.0.0.1";
...
+}
+#endif
+
+#if defined(OS_POSIX)
+static std::unique_ptr<Socket> CreateFDSocket(int fd) {
+  auto socket = ds2::make_unique<Socket>(fd);
+
+  errno = 0;
+  int flags = ::fcntl(fd, F_GETFL);
+  if (flags == -1 || errno == EBADF) {
```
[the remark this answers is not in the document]
yuex: yea, you're right, changed.

Sources/main.cpp
```diff
@@ -50,6 +50,40 @@ static std::string gDefaultHost = "127.0.0.1";
...
+    DS2LOG(Error, "cannot use fd %d: %s", fd, strerror(errno));
+  }
+
+  errno = 0;
```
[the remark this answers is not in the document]
yuex: changed

Sources/main.cpp
```diff
@@ -50,6 +50,40 @@ static std::string gDefaultHost = "127.0.0.1";
...
+  if (isatty(fd)) {
```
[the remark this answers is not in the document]
yuex: changed

Sources/main.cpp
yuex: But options are defined and parsed in the sub-mode's main functions. We don't know have the fd info in main(). I want to postpone the close only for g mode. That is why I changed the switch()

Sources/main.cpp
```diff
@@ -50,6 +50,40 @@ static std::string gDefaultHost = "127.0.0.1";
...
+    DS2LOG(Error, "cannot use fd %d: refers to a terminal", fd);
+  }
+
+  ::fcntl(fd, F_SETFD, FD_CLOEXEC);
```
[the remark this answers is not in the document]
yuex: Oh, I may need to add another public function to class Socket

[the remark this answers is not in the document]
sas: I don't think you need to add a public function. Just calling a private function from one of the constructors makes more sense I think.

Sources/main.cpp
sas: Good point. Let me think about this for a second.

Sources/main.cpp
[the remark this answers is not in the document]
yuex: but that may result in calling fcntl on an invalid fd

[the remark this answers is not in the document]
sas: I don't think so. You can do the validity checks first, like you do them right now, then when you create the `Socket` object, the constructor would set the CLOEXEC flag.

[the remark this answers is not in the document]
yuex: Ah, you are right

Sources/main.cpp
```diff
@@ -319,24 +356,47 @@ static int GdbserverMain(int argc, char **argv) {
 
   bool reverse = opts.getBool("reverse-connect");
 
+#if defined(OS_POSIX)
```
xiaobai: Move this block after creating a socket so you can minimize the number of ifdefs

Sources/main.cpp
xiaobai: I agree with this, you can just write these functions inline.

sas:
Gonna land this now. @xiaobai said he could take care of cleaning up the FD stuff.

PR #575, opened 2018-01-31
Add support for building Android from the NDK

xiaobai:
The purpose of this PR is to allow us to build Android using the android NDK instead of a standalone toolchain. We already run the tests with an NDK, building should be done that way too. Originally this PR was just to move us to CMake 3.7, but I now view that as only a part of the transition. Original comment left as-is.

Cmake 3.7 introduced support for cross compiling to Android out of the box.
This should help us build ds2 with the NDK more easily instead of introducing some nasty hackery to support as many API Levels as possible. CMake will take care of that for us now.


xiaobai:
Also, I want to see how the tests react to this. If our docker images don't have a recent cmake, we should work on that first.

xiaobai:
Also, I only upload patches I've tested myself. Even though our infra is failing on this, when we update docker images we should be good to go.

CMakeLists.txt
```diff
@@ -34,16 +34,16 @@ elseif (ARCH_NAME MATCHES "AMD64")
 endif ()
 
 if ("${OS_NAME}" MATCHES "Linux")
-  set(CMAKE_EXTRA_INCLUDE_FILES unistd.h sys/ptrace.h sys/uio.h sys/user.h)
+  set(CMAKE_EXTRA_INCLUDE_FILES unistd.h sys/ptrace.h sys/user.h)
```
sas: I don't think we need this if we specify the headers in the calls to `CHECK_SYMBOL_EXISTS`.

xiaobai: I think I agree here. I'll change it. :)

xiaobai:
Ok, I've added the last commit. We can now cross compile without the standalone toolchain. I plan on adding another commit to this PR removing the toolchain script and updating the documentation.

Something to note is that android now builds for the highest level API possible unless you specify a specific version. We can try to change that, but I kinda like it this way. ;P

xiaobai:
I pushed a docker image for this PR to the tag "2018-02-01", as that is the current date. I didn't update this PR pointing it to that tag (yet) so I can make sure I didn't accidentally push to latest. Il'l do that after.

xiaobai:
Per @sas's suggestion, the tag format will include the hour from now on. Should make life easy. :D

CMakeLists.txt
```diff
@@ -34,21 +34,22 @@ elseif (ARCH_NAME MATCHES "AMD64")
...
-  CHECK_FUNCTION_EXISTS(process_vm_writev HAVE_PROCESS_VM_WRITEV)
+  include(CheckSymbolExists)
+  set(CMAKE_REQUIRED_DEFINITIONS "-D_XOPEN_SOURCE=600")
+  CHECK_SYMBOL_EXISTS(posix_openpt "stdlib.h;fcntl.h" HAVE_POSIX_OPENPT)
+  CHECK_SYMBOL_EXISTS(gettid "unistd.h" HAVE_GETTID)
+  CHECK_SYMBOL_EXISTS(process_vm_readv "sys/uio.h" HAVE_PROCESS_VM_READV)
+  CHECK_SYMBOL_EXISTS(process_vm_writev "sys/uio.h" HAVE_PROCESS_VM_WRITEV)
+  set(CMAKE_REQUIRED_DEFINITIONS)
 
+  set(CMAKE_EXTRA_INCLUDE_FILES unistd.h sys/ptrace.h sys/user.h)
```
sas: Nit: put this after the `include()` so it matches the previous blocks (yes, OCD, I know.).

CMakeLists.txt
```diff
@@ -27,14 +27,13 @@ if (ARCH_NAME MATCHES "^ARM64" OR ARCH_NAME MATCHES "AARCH64")
   set(ARCH_NAME "ARM64")
 elseif (ARCH_NAME MATCHES "^ARM")
   set(ARCH_NAME "ARM") # Normalize ARMHF -> ARM
-elseif (ARCH_NAME MATCHES "I386")
+elseif (ARCH_NAME MATCHES "I386" OR ARCH_NAME MATCHES "I686")
```
sas: Nit: You could use `elseif (ARCH_NAME MATCHES "I[3-6]86")` so it matches other potential x86 architectures.

Support/Testing/Travis/install.py
```diff
@@ -62,9 +62,6 @@
 
 if target in android_toolchains:
     # Android builds get the toolchain from AOSP and use platform 21 by default.
-    android_platform = os.getenv('ANDROID_PLATFORM', "21")
-    check_call('./Support/Scripts/prepare-android-toolchain.sh "%s:%s"' %
```
sas: That's gonna break Travis for the android builds that haven't been moved to CircleCI yet, but I don't really care :p

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@
 
 docker_config: &docker_config
   docker:
-    - image: sas42/ds2-build-env
+    - image: sas42/ds2-build-env:2018-02-01_16
```
sas: Lol, by "hour" I meant something like `sas42/ds2-build-env:2018-02-01_16-51-42`.

xiaobai: I will change the tag name to 2018-02-01_17-00-00 when the clock strikes 5pm today.

CMakeLists.txt
xiaobai: Good idea, I'll do that.

CMakeLists.txt
xiaobai: mhm

Support/Testing/Travis/install.py
xiaobai: It'll break Travis next year, so who cares?

PR #576, opened 2018-02-01
Install a newer version of CMake on docker images

xiaobai:
Ubuntu's official repositories have CMake version 3.5.1, but we want version 3.7 or greater. This script and Dockerfile change should allow us to build docker images with the correct version of CMake moving forward.

After we get this in, I'll push new docker images.

### DIFF
--- a/Support/Testing/CircleCI/install-cmake.sh
+++ b/Support/Testing/CircleCI/install-cmake.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+##
+## Copyright (c) 2014-present, Facebook, Inc.
+## All rights reserved.
+##
+## This source code is licensed under the University of Illinois/NCSA Open
+## Source License found in the LICENSE file in the root directory of this
+## source tree. An additional grant of patent rights can be found in the
+## PATENTS file in the same directory.
+##
+
+cmake_version="3.7"
+cmake_package_name="cmake-${cmake_version}.0-Linux-x86_64"
+cmake_archive="${cmake_package_name}.tar.gz"
+if [ ! -e "${cmake_archive}" ] ; then
+  wget --continue --output-document="/tmp/${cmake_archive}" "https://cmake.org/files/v${cmake_version}/${cmake_archive}"
+fi
+
+tar --strip-components=1 -xf "/tmp/${cmake_archive}" -C "/usr/local"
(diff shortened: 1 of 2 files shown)


xiaobai:
I tested this by building a docker image with these changes, and verifying that it had CMake version 3.7.

Support/Testing/CircleCI/install-cmake.sh
```diff
@@ -0,0 +1,21 @@
...
+## This source code is licensed under the University of Illinois/NCSA Open
+## Source License found in the LICENSE file in the root directory of this
+## source tree. An additional grant of patent rights can be found in the
+## PATENTS file in the same directory.
+##
+
+cmake_version="3.7"
+cmake_package_name="cmake-${cmake_version}.0-Linux-x86_64"
+cmake_archive="${cmake_package_name}.tar.gz"
+cd "/usr/local/"
```
sas: Don't `cd` to `/usr/local`, download the archive in `/tmp` and just unpack it to `/usr/local` by specifying `-C` to `tar`. That way, it'll install the files to `/usr/local`, but the (now useless) archive is going to be deleted when you save the docker image.

xiaobai: Sounds good, updated to reflect that.

Support/Testing/CircleCI/install-cmake.sh
```diff
@@ -0,0 +1,19 @@
...
+## Source License found in the LICENSE file in the root directory of this
+## source tree. An additional grant of patent rights can be found in the
+## PATENTS file in the same directory.
+##
+
+cmake_version="3.7"
+cmake_package_name="cmake-${cmake_version}.0-Linux-x86_64"
+cmake_archive="${cmake_package_name}.tar.gz"
+if [ ! -e "${cmake_archive}" ] ; then
+  wget "https://cmake.org/files/v${cmake_version}/${cmake_archive}"
```
sas: Well that's not downloading to `/tmp`. :p

sas: I'd do this:

    wget --continue --output-document="/tmp/${cmake_archive}" "https://cmake.org/files/v${cmake_version}/${cmake_archive}"

xiaobai: Ahh, yeah. But the dockerfile moves the script to `/tmp` and executes it there. But making it explicit is probably better, yeah.

sas: Yeah but just because the script is in `/tmp` doesn't mean the output of `wget` will be located there. Your working directory doesn't change even if you run a script from another directory.

PR #577, opened 2018-02-06
Move Registers CI target to CircleCI

xiaobai:
Let's move this to CircleCI, seems easy enough.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,11 @@ test_step: &test_step
     name: Test
     command: cd build && ../Support/Scripts/run-lldb-tests.sh
 
+check_dirty_step: &check_dirty_step
+  run:
+    name: Check for dirty state
+    command: git --no-pager diff --exit-code
+
 build_only_defaults: &build_only_defaults
   <<: *docker_config
   steps:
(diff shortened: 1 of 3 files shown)


xiaobai:
Looks like Registers is failing because it can't identify a C++ compiler. Lol. I'll update the docker image tomorrow. 😛

.circleci/config.yml
```diff
@@ -55,6 +55,18 @@ jobs:
...
+  Registers:
+    <<: *docker_config
+    environment:
+      CLANG_FORMAT: clang-format-5.0
+    steps:
+      - checkout
+      - run:
+          name: Generate register descriptors
+          command: ./Support/Scripts/generate-register-descriptors.sh
+      - run:
```
sas: You can make this into a separate step, like we do for `build_step`, `configure_step`, etc, so you don't have to duplicate the code that's already in the `Style` check.

Support/Scripts/generate-register-descriptors.sh
```diff
@@ -23,7 +23,7 @@ if [ ! -x "$build_dir/regsgen2" ]; then
   echo "regsgen2 not found, building."
   mkdir -p "$build_dir"
   cd "$build_dir"
-  cmake ..
+  cmake .. -DCMAKE_TOOLCHAIN_FILE=../CMake/Toolchain-Linux-X86_64.cmake
```
sas: Put the `..` after the `-DCMAKE_...`.

.circleci/config.yml
```diff
@@ -43,6 +43,11 @@ build_and_test_defaults: &build_and_test_defaults
     - *build_step
     - *test_step
 
+check_dirty: &check_dirty
```
sas: Name this `check_dirty_step` and move it with the other build steps above. Also, remove the `- ` before `run` otherwise you won't be able to combine it with other steps.

.circleci/config.yml
```diff
@@ -52,9 +57,17 @@ jobs:
...
+    environment:
+      CLANG_FORMAT: clang-format-5.0
+    steps:
+      - checkout
       - run:
-          name: Check style differences
-          command: git --no-pager diff --exit-code
+          name: Generate register descriptors
+          command: ./Support/Scripts/generate-register-descriptors.sh
+      - <<: *check-dirty
```
sas: This should be `- *check_dirty_step`.

PR #578, opened 2018-02-07
Move the Documentation target from Travis to CircleCI

xiaobai:
Yeah, let's move it.


xiaobai:
Android-ARM is flaking :<

PR #579, opened 2018-02-07
Move Android API level 16 builds to CircleCI

xiaobai:
I created a new configure step. Although there's some repetition, I can't think of a better way to do this without introducing another layer of indirection (and I strongly prefer we don't introduce some wrapper script just to choose an API level properly).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ docker_config: &docker_config
 configure_step: &configure_step
   run:
     name: Configure
-    command: mkdir build && cd build && cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../Support/CMake/Toolchain-${TARGET-${CIRCLE_JOB}}.cmake ..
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -18,6 +18,11 @@ configure_step: &configure_step
     name: Configure
     command: mkdir build && cd build && cmake -GNinja -DCMAKE_TOOLCHAIN_FILE=../Support/CMake/Toolchain-${TARGET-${CIRCLE_JOB}}.cmake ..
 
+configure_android16_step: &configure_android16_step
```
sas: Instead of writing new rules for android API 16, you should probably add support for an env like "EXTRA_CMAKE_FLAGS" in `configure_step` so you don't have to duplicate the rules.

.circleci/config.yml
```diff
@@ -110,10 +110,20 @@ jobs:
     <<: *build_and_test_android
     environment:
       PLATFORM: 1
+  Android-ARM-16:
+    <<: *build_only_defaults
+    environment:
+      TARGET: Android-ARM
+      EXTRA_CMAKE_COMMANDS: "-DCMAKE_SYSTEM_VERSION=16"
```
sas: EXTRA_CMAKE_ARGS or EXTRA_CMAKE_FLAGS. These aren't really "commands".

PR #580, opened 2018-02-07
Move Linux-X86-Clang targets to CircleCI

xiaobai:
Let's move them :D

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,6 +99,13 @@ jobs:
     environment:
       TARGET: Linux-X86
       PLATFORM: 1
+  Linux-X86-Clang:
+    <<: *build_and_test_defaults
+  Linux-X86-Clang-platform:
+    <<: *build_and_test_defaults
+    environment:
+      TARGET: Linux-X86-Clang
+      PLATFORM: 1
   Linux-X86_64:
     <<: *build_and_test_defaults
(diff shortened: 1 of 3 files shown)


xiaobai:
I found 32bit libstdc++ in `/usr/lib32` on my docker image, we need to point it there probably.

.circleci/config.yml
```diff
@@ -90,6 +90,14 @@ jobs:
     environment:
       TARGET: Linux-X86
       PLATFORM: 1
+  Linux-X86-Clang:
+    <<: *build_and_test_defaults
+  Linux-X86-Clang-platform:
+    <<: *build_and_test_defaults
+    environment:
+      TARGET: Linux-X86-Clang
+      RELEASE: 1
```
sas: You can remove this line. We don't support `RELEASE=1` on CircleCI. I'm not sure if it's useful, but if it is, I'll add it later on.

xiaobai: Ah, alright. Sounds good.

xiaobai:
Yeah, I will move those in a separate PR.

xiaobai:
Once the effects of #584 kick in, I'll re-run the builds here and my expectation is that ds2 should build and the tests should run. I can't guarantee all tests will pass though. :P

PR #584, opened 2018-02-08
Move to GCC 5.4

xiaobai:
There are some stupid limitations around Clang and what it assumes about your system, as seen in #580. The easiest solution here is to just use the most up to date available toolchain and go from there.


Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,11 +40,11 @@ RUN apt-get install -y ninja-build
 RUN apt-get install -y flex bison
 
 # Install x86 compilers
-RUN apt-get install -y g++-4.9-multilib
+RUN apt-get install -y g++-multilib
 RUN apt-get install -y clang-5.0
 
 # Install arm compilers
-RUN apt-get install -y g++-4.9-multilib-arm-linux-gnueabi
```
sas: I don't know if you should touch these. The problem is just with multilib x86 vs. x86_64.

xiaobai: Surprisingly these are not the ones that break. :D

xiaobai: I think moving this is okay too, especially since it's practically free (in the sense that it's changing a few characters). It doesn't seem to break builds or tests (Android-ARM is actually running tests as I write this, but it's been flakey anyway). Do you feel strongly about keeping this at 4.9?

sas: No, I think you're right. I thought it might have been a mistake initially, but moving it makes sense.

PR #585, opened 2018-02-13
Move Documetation back to Travis for the time being

xiaobai:
It looks like we're going to have to talk to the Open Source team at Facebook to move this over properly. Until I figure out what needs to happen, let's keep this on Travis for now. I'd rather not remove our only way of updating documentation until we've got a solution on CircleCI.


.circleci/config.yml
```diff
@@ -83,15 +83,15 @@ jobs:
...
-  Documentation:
-    <<: *docker_config
-    environment:
-      ENCRYPTION_LABEL: c06a05533acc
-    steps:
-      - checkout
-      - run:
-          name: Generate documentation
-          command: ./Support/Scripts/generate-documentation.sh
+  #Documentation:
```
sas: I don't think you need to comment this block.

xiaobai: True! Let me uncomment it before I commit.

PR #586, opened 2018-02-13
Move gdb tests to CircleCI

xiaobai:
Let's move them!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@
 
 docker_config: &docker_config
   docker:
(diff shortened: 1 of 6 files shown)


xiaobai:
I want this to land after #585 so travis still has a target and doesn't fail. Ideally travis would be gone, but for reasons listed in that PR we can't quite do it yet.

sas:
You might wanna merge the top two commits of your branch. Otherwise, this is good to go.

xiaobai:
Yeah squashing them is a good idea I think.

PR #595, opened 2018-02-26
Pass log file to slave processes in platform mode

fjricci:
This makes --log-file actually useful for platform mode ds2 on a remote target, otherwise all the logs just get dumped into the void.

### DIFF
--- a/Headers/DebugServer2/Utils/Log.h
+++ b/Headers/DebugServer2/Utils/Log.h
@@ -34,7 +34,8 @@ enum {
 uint32_t GetLogLevel();
 void SetLogLevel(uint32_t level);
 void SetLogColorsEnabled(bool enabled);
-void SetLogOutputStream(FILE *stream);
+const char *GetLogOutputFilename();
+void SetLogOutputFilename(const char *filename);
 
 void Log(int level, char const *classname, char const *funcname,
          char const *format, ...) DS2_ATTRIBUTE_PRINTF(4, 5);
--- a/Sources/GDBRemote/PlatformSessionImpl.cpp
+++ b/Sources/GDBRemote/PlatformSessionImpl.cpp
@@ -111,6 +111,11 @@ ErrorCode PlatformSessionImplBase::onLaunchDebugServer(Session &session,
   } else if (GetLogLevel() == kLogLevelPacket) {
     args.push_back("--remote-debug");
   }
+  const char *outputFilename = GetLogOutputFilename();
+  if (outputFilename) {
+    args.push_back("--log-file");
+    args.push_back(outputFilename);
+  }
 #if defined(OS_POSIX)
   args.push_back("--setsid");
 #endif
--- a/Sources/Utils/Log.cpp
+++ b/Sources/Utils/Log.cpp
@@ -17,7 +17,16 @@
 #if defined(PLATFORM_ANDROID)
 #include <android/log.h>
 #endif
+#include <cstdio>
+#include <cstring>
+#include <fcntl.h>
+#include <limits.h>
 #include <sstream>
+#if !defined(OS_WIN32)
+#include <cerrno>
+#include <sys/stat.h>
+#include <unistd.h>
+#endif
 #include <vector>
 
 namespace {
@@ -31,6 +40,12 @@ FILE *sOutputStream = stdout;
 #else
 FILE *sOutputStream = stderr;
 #endif
+
+// PATH_MAX isn't defined when targeting MSVC
+#ifndef PATH_MAX
+#define PATH_MAX 1024
+#endif
+char sOutputFilename[PATH_MAX];
 } // namespace
 
 #if defined(PLATFORM_ANDROID)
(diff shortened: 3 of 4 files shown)


fjricci:
The concurrent bug only repros on circle and only sometimes and only in platform mode, so having platform mode logs would be very useful :)

Sources/Utils/Log.cpp
```diff
@@ -31,6 +40,12 @@ FILE *sOutputStream = stdout;
 #else
 FILE *sOutputStream = stderr;
 #endif
+
+// PATH_MAX isn't defined when targeting MSVC
+#ifndef PATH_MAX
+#define PATH_MAX 1024
```
xiaobai: Where is 1024 coming from? I'm assuming this is only for Windows platforms? Consider using `MAX_PATH` instead if that's the case.

Sources/Utils/Log.cpp
```diff
@@ -66,9 +81,28 @@ uint32_t GetLogLevel() { return sLogLevel; }
...
+  FILE *stream = fopen(filename, "a");
+  if (stream == nullptr) {
+    DS2LOG(Error, "unable to open %s for writing: %s", filename,
+           strerror(errno));
+  } else {
+#if defined(OS_POSIX)
+    // When ds2 is spawned by an app (e.g.: on Android), it will run with the
+    // app's user/group ID, and will create its log file owned by the app. By
+    // default, the permissions will be 0600 (rw-------) which makes us
+    // unable to get the log files. chmod() them to be able to access them.
```
xiaobai: Why does this apply to all POSIX OSes instead of just android? Would it happen on Linux or Tizen or any other POSIX platform we support? I know this isn't your code since you got it from `main.cpp`, but making it more specific couldn't hurt. :)

Sources/main.cpp
```diff
@@ -423,8 +408,6 @@ static int SlaveMain(int argc, char **argv) {
   opts.parse(argc, argv);
   HandleSharedOptions(opts);
 
-  ds2::SetLogLevel(ds2::kLogLevelWarning);
```
xiaobai: Why are you removing this?

Sources/GDBRemote/PlatformSessionImpl.cpp
```diff
@@ -111,6 +111,11 @@ ErrorCode PlatformSessionImplBase::onLaunchDebugServer(Session &session,
   } else if (GetLogLevel() == kLogLevelPacket) {
     args.push_back("--remote-debug");
   }
+  const char *outputFilename = GetLogOutputFilename();
+  if (outputFilename) {
```
xiaobai: Won't this will always be evaluated to true? `GetLogOutputFilename` will always give you a valid, non-null pointer since `sOutputFilename` is just a character array of length `PATH_MAX`. However, the actual contents of it could be {'\0', '\0', ..., '\0'} if it's not initialized. I don't think this is what you want.

Maybe something like:
```
if (strnlen(outputFilename, PATH_MAX) != 0)
```

Sources/Utils/Log.cpp
sas: This behavior depends on what the process' `umask` is. By default, it'll be `022` which means the log file will be created with mode `755` but on Android the umask happens to be `077` sometimes. Always forcing the file to be readable makes things a little bit easier.

That being said, if `umask` is *not* `022` on Linux, it's probably set by the user for a reason, so making this `fchmod()` only on Android could be a better behavior.

Headers/DebugServer2/Utils/Log.h
```diff
@@ -34,7 +34,8 @@ enum {
 uint32_t GetLogLevel();
 void SetLogLevel(uint32_t level);
 void SetLogColorsEnabled(bool enabled);
-void SetLogOutputStream(FILE *stream);
+const char *GetLogOutputFilename();
+void SetLogOutputFilename(const char *filename);
```
sas: I think you should change these APIs to use `std::string()` instead of `char *`. It'll make the implementation simpler. It'll also conveniently allow you to ignore this `PATH_MAX` thing.

Sources/GDBRemote/PlatformSessionImpl.cpp
sas: Good catch by @xiaobai. Your storage for the output file name is a statically allocated `char` array, so you'll never return a `NULL` pointer from `GetLogOutputFilename()`. If you switch to using `std::string` in `Log.cpp`, you can just check for `GetLogOutputFilename().length() > 0`.

Sources/main.cpp
sas: Yeah, if we have a log file we want all the debug logs to go there, forcing to warning will suppress these logs, so I think removing this line makes sense.

sas:
I think I'm just gonna grab this PR and make the modifications I listed before landing it.

fjricci:
Thanks

sas:
Moved your commits to PR#603.

PR #597, opened 2018-02-26
Enable Concurrent testing

fjricci:
Depends on enable/disable PR, but putting up separately in case there are other issues.


xiaobai:
Looks like there's some overlap with #523?

fjricci:
Yes, intentionally so (see comment)

xiaobai:
Ah I see, didn't know what you meant by "Depends on enable/disable PR". Thanks for clarifying.

fjricci:
Yeah, I was going to link the PR number but I was lazy 😆

fjricci:
rebase --ignore-date doesn't seem to have totally worked here, some of the commits from https://github.com/facebook/ds2/pull/523 are not in order with respect to the ESRCH commits. Should be resolved once that PR goes in though.

xiaobai:
Yeah, github shows commits in linear order based on commit time. It definitely commits them in the correct order though.

fjricci:
Damn circleci isn't fooled by close-open

lanza:
Rebased and updated in https://github.com/facebook/ds2/pull/644

PR #601, opened 2018-03-09
Update instructions for running docker image locally

xiaobai:
In order to more closely match CircleCI locally, we should be running the docker image slightly differently. Through experimentation and reading some discussions online, it looks like CircleCI runs the docker images in a more privileged mode. The instructions should try to match that.

### DIFF
--- a/Support/Testing/CircleCI/Dockerfile
+++ b/Support/Testing/CircleCI/Dockerfile
@@ -17,7 +17,9 @@
 #
 # This hash identifies the image you just built. You can then run it locally
 # with the following command:
-#     docker run --rm -it --name ds2-testing 360abf8e6246
+#     docker run --privileged --security-opt seccomp:unconfined --rm -it --name ds2-testing 360abf8e6246
+# We run in privileged mode so that we may use ptrace to its full
+# extent as needed.
 #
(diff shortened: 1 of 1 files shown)


Support/Testing/CircleCI/Dockerfile
```diff
@@ -17,7 +17,9 @@
 #
 # This hash identifies the image you just built. You can then run it locally
 # with the following command:
-#     docker run --rm -it --name ds2-testing 360abf8e6246
+#     docker run --privileged --security-opt seccomp:unconfined --rm -it --name ds2-testing 360abf8e6246
```
sas: Do you need both `--privileged` and `--security-opt seccomp:unconfied`? I think I only needed the latter.

Also, are these the flags with which CircleCI runs our images?

xiaobai: Yes, I believe these are the same flags.

I'm not sure if we need both flags. I was under the impression we need both, as they enable different capabilities.

sas:
I think only `--security-opt` is required, but I'm not gonna ask you to
test that :p Feel free to merge this.
-- 
-- 
Stephane Sezer

PR #602, opened 2018-03-23
Remove dead patches

xiaobai:
These patches have already been committed upstream and we no longer try to even apply them. Let's delete them.


fjricci:
❤️

PR #606, opened 2018-04-04
Rename Type in BreakpointManager to Lifetime

xiaobai:
It's definitely true you can use ask what type of breakpoint you have, and get back an answer concerning its lifetime (Temporary, Permanent, etc). However, we also have BreakpointType describing whether or not its a software breakpoint, hardware breakpoint, what kind of watchpoint, etc.

I think `Lifetime` is a much clearer name to describe what this is supposed to represent.

### DIFF
--- a/Sources/Core/BreakpointManager.cpp
+++ b/Sources/Core/BreakpointManager.cpp
@@ -22,8 +22,8 @@ BreakpointManager::~BreakpointManager() {
 
 void BreakpointManager::clear() { _sites.clear(); }
 
-ErrorCode BreakpointManager::add(Address const &address, Type type, size_t size,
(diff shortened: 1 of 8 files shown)


Sources/Core/BreakpointManager.cpp
```diff
@@ -159,11 +160,11 @@ bool BreakpointManager::hit(Address const &address, Site &site) {
     return false;
 
   //
-  // If this breakpoint type becomes 0, it will be erased after calling
+  // If this breakpoint's lifetime is set to 0, it will be erased after calling
```
sas: Nano-nit: "becomes" makes more sense that "is set to" in this case I think.

PR #607, opened 2018-04-05
Remove assertions and add debug logs to breakpoint validation code

xiaobai:
We shouldn't assert (and consequently die) when validating user input. Rather, we should just return an error and add some debug logs that lets us know that's happening.

I went through and added some logs to other breakpoint validation code as well.

### DIFF
--- a/Sources/Core/ARM/SoftwareBreakpointManager.cpp
+++ b/Sources/Core/ARM/SoftwareBreakpointManager.cpp
@@ -171,7 +171,10 @@ void SoftwareBreakpointManager::getOpcode(uint32_t type,
 ErrorCode SoftwareBreakpointManager::isValid(Address const &address,
                                              size_t size, Mode mode) const {
   DS2ASSERT(mode == kModeExec);
-  DS2ASSERT((size >= 2) && (size <= 4));
+  if (size < 2 || size > 4) {
+    DS2LOG(Debug, "Received unsupported breakpoint size %zu", size);
+    return kErrorInvalidArgument;
+  }
 
   return super::isValid(address, size, mode);
 }
--- a/Sources/Core/X86/HardwareBreakpointManager.cpp
+++ b/Sources/Core/X86/HardwareBreakpointManager.cpp
@@ -214,14 +214,18 @@ ErrorCode HardwareBreakpointManager::isValid(Address const &address,
 
   case 2:
   case 4:
-    if (mode == kModeExec)
+    if (mode == kModeExec) {
       return kErrorInvalidArgument;
+    }
     break;
   default:
+    DS2LOG(Debug, "Received unsupported hardware breakpoint size %zu", size);
     return kErrorInvalidArgument;
   }
 
   if ((mode & kModeExec) && (mode & (kModeRead | kModeWrite))) {
+    DS2LOG(Debug, "Trying to set a hardware breakpoint with mixed exec and "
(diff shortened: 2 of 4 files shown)


Sources/Core/ARM/SoftwareBreakpointManager.cpp
```diff
@@ -170,8 +170,15 @@ void SoftwareBreakpointManager::getOpcode(uint32_t type,
...
 ErrorCode SoftwareBreakpointManager::isValid(Address const &address,
                                              size_t size, Mode mode) const {
-  DS2ASSERT(mode == kModeExec);
-  DS2ASSERT((size >= 2) && (size <= 4));
+  if (size < 2 || size > 4) {
+    DS2LOG(Debug, "Received unsupported breakpoint size %zu", size);
+    return kErrorInvalidArgument;
+  }
+
+  if (mode != kModeExec) {
```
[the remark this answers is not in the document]
xiaobai: Good point.

Sources/Core/ARM/SoftwareBreakpointManager.cpp
```diff
@@ -170,8 +170,12 @@ void SoftwareBreakpointManager::getOpcode(uint32_t type,
 
 ErrorCode SoftwareBreakpointManager::isValid(Address const &address,
                                              size_t size, Mode mode) const {
+  if (size < 2 || size > 4) {
+    DS2LOG(Debug, "Received unsupported breakpoint size %zu", size);
+    return kErrorInvalidArgument;
+  }
+
   DS2ASSERT(mode == kModeExec);
```
sas: You should keep these assertions at the top of the function.

Sources/Core/X86/SoftwareBreakpointManager.cpp
```diff
@@ -60,7 +60,11 @@ void SoftwareBreakpointManager::getOpcode(uint32_t type,
 
 ErrorCode SoftwareBreakpointManager::isValid(Address const &address,
                                              size_t size, Mode mode) const {
-  DS2ASSERT(size == 0 || size == 1);
+  if (size != 0 && size != 1) {
+    DS2LOG(Debug, "Received unsupported breakpoint size %zu", size);
+    return kErrorInvalidArgument;
+  }
+
   DS2ASSERT(mode == kModeExec);
```
sas: Assertions (pre-condition checks) should be at the top of the function.

Sources/Core/X86/HardwareBreakpointManager.cpp
```diff
@@ -214,14 +214,16 @@ ErrorCode HardwareBreakpointManager::isValid(Address const &address,
 
   case 2:
   case 4:
-    if (mode == kModeExec)
-      return kErrorInvalidArgument;
+    DS2ASSERT(mode != kModeExec);
```
sas: This is user input, it shouldn't be an assert.

Sources/Core/X86/HardwareBreakpointManager.cpp
```diff
@@ -214,14 +214,18 @@ ErrorCode HardwareBreakpointManager::isValid(Address const &address,
 
   case 2:
   case 4:
-    if (mode == kModeExec)
+    if (mode == kModeExec) {
```
sas: Nice.

PR #608, opened 2018-04-10
Simplify compiler support code

sas:


Headers/DebugServer2/Utils/CompilerSupport.h
```diff
@@ -12,26 +12,7 @@
...
-#if !defined(__has_builtin)
-#if defined(COMPILER_GCC)
-#define __has_builtin(BUILTIN) 1
-#else
-#define __has_builtin(BUILTIN) 0
-#endif
-#endif
-
-#if __has_attribute(format)
+#if defined(COMPILER_GCC) || defined(COMPILER_CLANG)
```
xiaobai: `#endif` is where?

Headers/DebugServer2/Utils/CompilerSupport.h
```diff
@@ -66,23 +41,19 @@
...
-#error "Compiler attribute __packed__ is required."
 #endif
 
 #if defined(COMPILER_MSVC)
 #define DS2_ATTRIBUTE_ALIGNED "DS2_ATTRIBUTE_ALIGNED not implemented on MSVC"
-#elif __has_attribute(__aligned__)
+#elif defined(COMPILER_GCC) || defined(COMPILER_CLANG)
 #define DS2_ATTRIBUTE_ALIGNED(SIZE) __attribute__((__aligned__(SIZE)))
-#else
-#error "Compiler attribute __aligned__ is required."
```
xiaobai: Might be worth keeping some of these `#errors` around, at least to tell the user that their compiler isn't supported.

Headers/DebugServer2/Utils/CompilerSupport.h
xiaobai: Wait, this becomes only for `__has_attribute(format)` now right?

Headers/DebugServer2/Utils/CompilerSupport.h
sas: If the compiler doesn't support it, you'll get errors during the build anyway.

Headers/DebugServer2/Utils/CompilerSupport.h
sas: I don't understand the question :p

xiaobai: I misunderstood what this was replacing. It looks fine to me now.

Headers/DebugServer2/Utils/CompilerSupport.h
xiaobai: True.

xiaobai:
Your test failure is related to watchpoints, which I bet is a flake. If you retry it, it'll probably pass.

PR #609, opened 2018-04-12
Add support for choosing breakpoint sizes

xiaobai:
According to the gdb-remote protocol, if the debugger does not choose an
architecture-specific size for a breakpoint, it is valid for it to send
0 as the breakpoint size. This value indicates that the debugserver
should choose a breakpoint size.

I had a workaround in #600 that I've now removed and replaced with this commit. If the all the related Linux-x86{_64,} tests pass there, this should be good to go in.

### DIFF
--- a/Headers/DebugServer2/Core/BreakpointManager.h
+++ b/Headers/DebugServer2/Core/BreakpointManager.h
@@ -81,6 +81,7 @@ public:
 protected:
   virtual ErrorCode isValid(Address const &address, size_t size,
                             Mode mode) const;
+  virtual size_t chooseBreakpointSize() const = 0;
 
 protected:
   friend Target::ProcessBase;
--- a/Sources/Core/ARM/HardwareBreakpointManager.cpp
+++ b/Sources/Core/ARM/HardwareBreakpointManager.cpp
@@ -40,4 +40,9 @@ ErrorCode HardwareBreakpointManager::isValid(Address const &address,
   DS2LOG(Debug, "Trying to set hardware breakpoint on arm");
   return kErrorUnsupported;
 }
+
+size_t HardwareBreakpointManager::chooseBreakpointSize() const {
+  DS2BUG(
+      "Choosing a hardware breakpoint size on ARM is an unsupported operation");
+}
 } // namespace ds2
--- a/Sources/Core/ARM/SoftwareBreakpointManager.cpp
+++ b/Sources/Core/ARM/SoftwareBreakpointManager.cpp
@@ -178,4 +178,9 @@ ErrorCode SoftwareBreakpointManager::isValid(Address const &address,
 
   return super::isValid(address, size, mode);
 }
+
+size_t SoftwareBreakpointManager::chooseBreakpointSize() const {
+  DS2BUG(
+      "Choosing a software breakpoint size on ARM is an unsupported operation");
+}
 } // namespace ds2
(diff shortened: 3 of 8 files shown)


xiaobai:
Looks like it passes on #600. This is ready for review.

Headers/DebugServer2/Core/BreakpointManager.h
```diff
@@ -82,6 +82,8 @@ class BreakpointManager {
   virtual ErrorCode isValid(Address const &address, size_t size,
                             Mode mode) const;
```
sas: you should remove this blank line.

Sources/Core/ARM/HardwareBreakpointManager.cpp
```diff
@@ -40,4 +40,9 @@ ErrorCode HardwareBreakpointManager::isValid(Address const &address,
   DS2LOG(Debug, "Trying to set hardware breakpoint on arm");
   return kErrorUnsupported;
 }
+
+size_t HardwareBreakpointManager::chooseBreakpointSize() {
+  DS2BUG(
```
sas: `DS2BUG()` is like an assert that stays even in release builds, so I think we shouldn't do that. We should somehow return an error to the caller.

Sources/Core/ARM/SoftwareBreakpointManager.cpp
```diff
@@ -178,4 +178,9 @@ ErrorCode SoftwareBreakpointManager::isValid(Address const &address,
 
   return super::isValid(address, size, mode);
 }
+
+size_t SoftwareBreakpointManager::chooseBreakpointSize() {
+  DS2BUG(
```
sas: Ditto.

Sources/Core/X86/HardwareBreakpointManager.cpp
```diff
@@ -236,6 +236,11 @@ ErrorCode HardwareBreakpointManager::isValid(Address const &address,
   return super::isValid(address, size, mode);
 }
 
+size_t HardwareBreakpointManager::chooseBreakpointSize() {
+  DS2BUG(
```
sas: Ditto.

Sources/Core/X86/SoftwareBreakpointManager.cpp
```diff
@@ -51,10 +51,7 @@ int SoftwareBreakpointManager::hit(Target::Thread *thread, Site &site) {
 
 void SoftwareBreakpointManager::getOpcode(uint32_t type,
                                           ByteVector &opcode) const {
-  DS2ASSERT(type == 1);
```
sas: I might be missing something, but I think this assertion should stay here. If everything goes as expected, either we were asked to set a breakpoint of size 1, or we were sent 0, and we chose a breakpoint of size 1 in `chooseBreakpointSize()`, or we were sent some other value, and we should have returned an error to the debugger and *not* added the breakpoint. So here, type should *always* be `1`.

Sources/Core/ARM/HardwareBreakpointManager.cpp
xiaobai: How do you think we should return an error to the caller? DS2BUG is a fatal assert indicative of a bug. Maybe the whole "This is FATAL, cannot recover" is a bit extreme. Maybe a warning?

Sources/Core/X86/SoftwareBreakpointManager.cpp
xiaobai: True.

PR #610, opened 2018-04-16
Improve run-lldb-tests.sh

sas:


xiaobai:
Am I correct in assuming this just lets us test against lldb-server in a docker image quickly?

sas:
> Am I correct in assuming this just lets us test against lldb-server in a docker image quickly?

Basically, yeah.

PR #612, opened 2018-05-01
Fix x86 and x86_64 floating point register support on Linux

xiaobai:
I understand this is a big change, and will probably require a lot of time to digest. I'm sure there are a few improvements to be made in implementation, but I think the fundamental idea is correct.

I refactored this in the process of fixing TestRegisters. I believe our previous implementation of dealing with floating point registers was incorrect. I have refactored the code to fix at least some of the problems, but not all. I think I have fixed the following:
- fiseg/fioff/foseg/fooff were incorrectly aliased on x86_64. Especially strange given fiseg/foseg were aliased to cs.
- No support for xstate_header, and therefore no way of enabling features (e.g. x87, sse, avx) through the features mask.
- Incorrect read/writes to stmm registers. Specifically, we did not account for 6 bytes of padding between each stmm register in user_fpxregs_struct. This meant that we were reading from and writing to buffer space.
- user_fpxregs_struct fields are now defined with explicit sizes.
- Not handling xcr0 at all, meaning we never could tell what features were actually enabled.

There are a few things I did not fix, but I added TODOs to revisit. These are not especially critical and would make for great beginner tasks for ds2.
- Support xcomp_bv field in xstate_header.
- Fall back to FXSAVE when XSAVE fails.

### DIFF
--- a/Headers/DebugServer2/Architecture/X86/CPUState.h
+++ b/Headers/DebugServer2/Architecture/X86/CPUState.h
@@ -46,6 +46,20 @@ struct X87Register {
   };
 };
 
+enum XFeature : uint64_t {
+  X86_X87 = 1 << 0,
+  X86_SSE = 1 << 1,
+  X86_AVX = 1 << 2,
+  X86_MPX_BNDREGS = 1 << 3,
+  X86_MPX_CSR = 1 << 4,
+  X86_AVX_512_OPMASK = 1 << 5,
+  X86_AVX_512_HI256 = 1 << 6,
+  X86_AVX_512_ZMM = 1 << 7,
+  X86_PROC_TRACE = 1 << 8,
+  X86_PROT_KEYS_USER_REGS = 1 << 9,
+  X86_UNKNOWN_XFEATURE = 1 << 10,
+};
+
 struct CPUState {
   union {
     uint32_t regs[16];
@@ -91,7 +105,9 @@ struct CPUState {
     } avx;
   };
 
-  // TODO - add information about xstate_hdr here
+  struct {
+    uint64_t xfeatures_mask;
+  } xsave_header;
 
   struct {
     uint32_t dr[8];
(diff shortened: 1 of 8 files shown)


xiaobai:
It looks like the Linux failures are unrelated to floating point register support. The android-arm failures are the same as always, the test suite just doesnt run correctly. :)

Headers/DebugServer2/Architecture/X86_64/CPUState.h
```diff
@@ -85,8 +86,18 @@ struct CPUState64 {
...
+      struct {
+        uint64_t firip;
+        uint64_t firdp;
+      } x86_64;
+      struct {
+        uint32_t fioff;
+        uint32_t fiseg;
+        uint32_t fooff;
+        uint32_t foseg;
+      } x86;
```
xiaobai: It's more useful to expose as 4 separate registers instead of as 2 registers here. LLDB expects these 4 registers instead of firip and firdp anyway.

Headers/DebugServer2/Architecture/X86/CPUState.h
```diff
@@ -33,17 +33,22 @@ struct AVXVector {
...
-    // https://msdn.microsoft.com/en-us/library/9cx8xs15.aspx
-    long double f80;
-#endif
-    double f64;
-    float f32;
-    uint64_t mm;
-    uint8_t bytes[10];
-  };
+  uint8_t data[10];
+  uint8_t padding[6];
```
xiaobai: Padding exposes implementation details. Remove it and create a constant with the amount of padding that can be used instead.

Headers/DebugServer2/Host/Linux/ExtraWrappers.h
```diff
@@ -31,6 +31,8 @@
 #endif // ARCH_ARM && !ARM_VFPREGS_SIZE
 
 #if defined(ARCH_X86) || defined(ARCH_X86_64)
+const size_t x87Padding = 6;
```
sas: You should define this where you use it. It's not something we want to export to other parts of the code.
Also, you can make it `static constexpr` instead of `const`.

xiaobai: It will be defined in a few places then, which I don't find nicer than this at all. However, `static constexpr` does sound nicer :D

sas: Ah you're right. I didn't see how much it was used.

PR #614, opened 2018-05-03
Re-enable TestIntegerTypes

xiaobai:
This test was disabled as a part of the move to LLVM 6.0.


xiaobai:
Closing, this isn't gonna magically work. in fact, TestFloatTypes is flaking again too. :)

PR #618, opened 2018-05-14
Refactor support for using lldb-server in run-lldb-tests.sh

xiaobai:
I did some refactoring, because it's not enough to just not set LLDB_DEBUGSERVER_PATH. instead, the idea is to set it to lldb-server-6.0.

I didn't add support for using lldb-server and the `--log` flag at the same time (since lldb-server does logging differently than ds2) but I bet it wouldn't be too hard to implement if we cared to see what lldb-server does.

I've run both the test suite with both ds2 and lldb-server. Surprisingly lldb-server has more failures, but the test suite ran on both.

Additionally I shuffled some code around.

### DIFF
--- a/Support/Scripts/run-lldb-tests.sh
+++ b/Support/Scripts/run-lldb-tests.sh
@@ -230,7 +230,7 @@ if [[ "${PLATFORM-}" = "1" ]]; then
     args+=("--excluded" "$blacklist_dir/ds2/platform.blacklist")
   fi
 
-  server_args=("p" "--server" "--listen" "$server_port")
+  server_args=("p" "--server" "--listen" "*:$server_port")
 
   if ! $opt_use_lldb_server && $opt_log; then
     server_args+=("--remote-debug" "--log-file=$working_dir/$(basename "$server_path").log")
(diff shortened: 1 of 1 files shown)


Support/Scripts/run-lldb-tests.sh
```diff
@@ -226,7 +226,7 @@ if [[ "${PLATFORM-}" = "1" ]]; then
     args+=("--excluded" "$blacklist_dir/ds2/platform.blacklist")
   fi
 
-  server_args=("p" "--server" "--listen" "$server_port")
```
sas: I think I remember this. lldb-server does not need the `*:` in gdbserver mode but does need them in platform mode, which is a little bit confusing.

sas:
Also, I don't understand this:

> it's not enough to just not set LLDB_DEBUGSERVER_PATH. instead, the idea is to set it to lldb-server-6.0

Why is it not enough? What happens when we don't set `$LLDB_DEBUGSERVER_PATH` at all? My expectation would be that lldb will find the newly build `lldb-server` binary (i.e.: `lldb-server-6.0` here) and just use that. What does it do instead?

xiaobai:
ping @sas 
any other problems here? Otherwise I think this is good to go in.

PR #622, opened 2018-05-18
Re-enable TestReturnValue except on Linux-x86

xiaobai:
I got this to pass locally on my Docker image (Linux-X86_64, didn't check platform mode). If CircleCI is happy with it, we should re-enable it.


xiaobai:
Seems like Linux-X86 doesn't appreciate this, so I'll debug it on there. Hopefully we'll be able to re-enable this test.

xiaobai:
I updated this PR to still blacklist TestReturnValue on x86 (where it seems to be expected to fail anyway). It passes when you use lldb-server, so I blacklisted it only with ds2.

PR #624, opened 2018-05-22
Re-enable TestEnumTypes

xiaobai:
I'm going to just see how CircleCI reacts. It passed on Linux-X86_64 locally.


xiaobai:
Looks like it passes! Aside from Android-ARM being totally broken, that is.

PR #627, opened 2018-05-29
Update Linux::Process to use Linux::PTrace instead of POSIX::PTrace

lanza:
`Linux::Process` holds a `Linux::PTrace` yet has a method that returns a `POSIX::PTrace`.

Change `Linux::Process::PTrace() const` to return a `Linux::PTrace` instead of a `POSIX::PTrace`.


lanza:
This fails due to spurious dependencies from `Linux::Process` on `POSIX::PTrace` and NOT `Linux::PTrace`.

PR #628, opened 2018-05-29
Implement Linux::PTrace::getClonedTID(ProcessID, unsigned long&) method to be used to call ptrace(3) with PTRACE_GETEVENTMSG in the case of a tracee clone(3) call

lanza:
Upon a tracee calling `clone(3)`, the tracer can call `ptrace(3)` with the `PTRACE_GETEVENTMSG` flag to receive the `tid` of the newly cloned thread.

Implement a method, `Linux::PTrace::getClonedTID` to be called after a clone event that sets the value of `clonedTID` to the newly cloned child's `tid`.


lanza:
@sas @xiaobai This should work now.

lanza:
Closing this and just going to use the one PR.

PR #630, opened 2018-05-30
Modify Linux::Process::wait() to account for a race condition involving a simultaneous clone & step

lanza:
It is possible for the inferior to experience a (roughly) simultaneous
SIGTRAP from stepping while also receiving a SIGTRAP/SIGSTOP pair from a
clone(3) call. Depending on the receiving order of the signals, ds2
could handle the situation improperly and miss out on the existance of
the cloned thread.

Modify the processing logic to check if we are simultaneously
experiencing a thread being stepped while that thread also calls
clone and, if so, handle the cloned child at within the function instead
of depending on the normal waitpid loop.

### DIFF
--- a/Headers/DebugServer2/Host/Linux/PTrace.h
+++ b/Headers/DebugServer2/Host/Linux/PTrace.h
@@ -66,6 +66,7 @@ public:
 
 public:
   ErrorCode getSigInfo(ProcessThreadId const &ptid, siginfo_t &si) override;
+  ErrorCode getEventMessage(ProcessThreadId const &ptid, unsigned long &data);
 
 protected:
   virtual ErrorCode readRegisterSet(ProcessThreadId const &ptid, int regSetCode,
--- a/Sources/Host/Linux/PTrace.cpp
+++ b/Sources/Host/Linux/PTrace.cpp
@@ -285,6 +285,17 @@ ErrorCode PTrace::getSigInfo(ProcessThreadId const &ptid, siginfo_t &si) {
   return kSuccess;
 }
 
+ErrorCode PTrace::getEventMessage(ProcessThreadId const &ptid,
+                                  unsigned long &data) {
+  pid_t pid;
+  CHK(ptidToPid(ptid, pid));
+
+  if (wrapPtrace(PTRACE_GETEVENTMSG, pid, nullptr, &data) < 0)
+    return Platform::TranslateError();
+
+  return kSuccess;
+}
+
 ErrorCode PTrace::readRegisterSet(ProcessThreadId const &ptid, int regSetCode,
                                   void *buffer, size_t length) {
   struct iovec iov = {buffer, length};
--- a/Sources/Target/Linux/Process.cpp
+++ b/Sources/Target/Linux/Process.cpp
@@ -215,17 +215,55 @@ ErrorCode Process::wait() {
 
     switch (_currentThread->_stopInfo.event) {
     case StopInfo::kEventNone:
-      // If the thread is stopped for no reason, it means the debugger (ds2)
-      // sent a SIGSTOP to it while it was already stopped for another reason,
-      // the SIGSTOP was queued, and now we're getting this notification. We
-      // have to ignore this and just let the thread continue.
-      // If `stepping` is true, it means the thread was actually being
-      // single-stepped before stopping, so instead of doing a `resume()`, we
-      // have to do a new `step()`.
-      if (stepping) {
+      // If _stopInfo.event is kEventNone it means the thread is either
+      // 1. stopped because another thread of the inferior stopped (e.g. because
+      //    of a breakpoint or received a signal, etc) and ds2 sent a SIGSTOP
+      //    to pause this thread while the other thread's stop is evaluated. In
+      //    this case we just resume the thread.
+      // 2. stopped because this thread called clone(2) (or similar e.g.
+      //    pthread_create). In this case we have two subcases:
+      //    a. The thread was running and hit the clone/spawn and thus was
+      //       stopped with a SIGTRAP. We just resume it and goto
+      //       continue_waiting. The loop will continue and the next waitpid
+      //       will handle the freshly spawned thread. In this case we call
+      //       resume as well.
+      //    b. We were single stepping (c.f. the `stepping` var) and
+      //       were stopped by a SIGTRAP for the step while simultaneously
+      //       being stopped via the SIGTRAP for the clone/spawn.
+      //       In this case we can run into a race condition with the
+      //       waitpid(3) call and thus have to handle the cloned thread
+      //       and the currentThread at once. So we call step on the
+      //       _currentThread and then beforeResume the cloned thread.
+      //
+      // We also have another theoretically possible case where we are not
+      // stopped by a clone but still stepping while encountering a kEventNone.
+      // This should not happen and we just assert that it does not.
+
+      if (_currentThread->_stopInfo.reason == StopInfo::kReasonThreadSpawn &&
+          stepping) { //(2b)
+        unsigned long spawnedThreadIdData;
+        CHK(ptrace().getEventMessage(_pid, spawnedThreadIdData));
+        auto const spawnedThreadId = static_cast<ThreadId>(spawnedThreadIdData);
+
+        int spawnedThreadStatus;
+        ThreadId const returnedThreadId =
+            ::waitpid(spawnedThreadId, &spawnedThreadStatus, WNOHANG | __WALL);
+        if (spawnedThreadId != returnedThreadId)
+          return kErrorProcessNotFound;
+        DS2LOG(Debug, "child thread tid %d %s", returnedThreadId,
+               Stringify::WaitStatus(spawnedThreadStatus));
+
         _currentThread->step();
+        auto newThread = new Thread(this, returnedThreadId);
+        newThread->beforeResume();
+      } else if (stepping) {
+        // We should never see a case where we're:
+        //   1. stopped for event kEventNone
+        //   2. stepping
+        //   3. not stopped for reason kReasonThreadSpawn
+        DS2BUG("inconsistent thread stop info");
       } else {
-        _currentThread->resume();
+        _currentThread->resume(); // (1) and (2a)
       }
       goto continue_waiting;
 
--- a/Support/Testing/Blacklists/ds2/x86_64.blacklist
+++ b/Support/Testing/Blacklists/ds2/x86_64.blacklist
@@ -1,2 +1,0 @@
-xfail
-TestCreateDuringInstructionStep.CreateDuringInstructionStepTestCase.test_step_inst
--- a/Sources/Target/Linux/Thread.cpp
+++ b/Sources/Target/Linux/Thread.cpp
@@ -84,6 +84,7 @@ ErrorCode Thread::updateStopInfo(int waitStatus) {
 
     if (waitStatus >> 8 == (SIGTRAP | (PTRACE_EVENT_CLONE << 8))) { // (1)
       _stopInfo.event = StopInfo::kEventNone;
+      _stopInfo.reason = StopInfo::kReasonThreadSpawn;
     } else if (si.si_code == SI_TKILL && si.si_pid == getpid()) { // (2)
       // The only signal we are supposed to send to the inferior is a SIGSTOP.
       DS2ASSERT(_stopInfo.signal == SIGSTOP);
--- a/Support/Scripts/run-lldb-tests.sh
+++ b/Support/Scripts/run-lldb-tests.sh
@@ -163,9 +163,6 @@ fi
 if [ -n "${TARGET-}" ]; then
   if [[ "${TARGET}" == "Linux-X86_64" || "${TARGET}" == "Linux-X86_64-Clang" ]]; then
     args+=("--arch=x86_64")
-    if ! $opt_no_ds2_blacklists; then
-      args+=("--excluded" "$blacklist_dir/ds2/x86_64.blacklist")
-    fi
     if [[ "${PLATFORM-}" = "1" ]]; then
       args+=("--excluded" "$blacklist_dir/ds2/llvm_60.blacklist")
     fi


Sources/Host/POSIX/PTrace.cpp
```diff
@@ -125,6 +125,25 @@ ErrorCode PTrace::resume(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
   return doStepResume(false, ptid, signal, address);
 }
 
+ErrorCode PTrace::getClonedTID(ProcessId pid, unsigned long &clonedTID) {
```
sas: This is really sketchy because `PTRACE_GETEVENTMSG` is used for more than getting just the clone TID. You might wanna name it `getEventMessage()` and let the caller interpret the result (because the caller knows what the last ptrace event was and can interpret the result correctly).

Sources/Host/POSIX/PTrace.cpp
```diff
@@ -125,6 +125,25 @@ ErrorCode PTrace::resume(ProcessThreadId const &ptid, ProcessInfo const &pinfo,
...
+    return Platform::TranslateError();
+  }
+
+  return kSuccess;
+#elif defined(OS_FREEBSD) || defined(OS_DARWIN)
+  // PTRACE_GETEVENTMSG does not exist on Darwin. TODO: Find alternative
+  // implementation. Alternatively, this method could be in Linux::PTrace, but
+  // various bugs exist
+  //   that cause Linux::Process to depend on it's Linux::Process::ptrace()
+  //   method returning a POSIX::PTrace and NOT a Linux::PTrace
```
sas: I'm not sure what the problem is here, you can move this function to `Host::Linux::PTrace`, and make `Target::Linux::PTrace` return a `Host::Linux::PTrace` instead of a `Host::POSIX::PTrace`. See PR#635.

Headers/DebugServer2/Types.h
```diff
@@ -146,6 +146,7 @@ struct StopInfo {
     kReasonThreadSpawn,
     kReasonThreadEntry,
     kReasonThreadExit,
+    kReasonClone,
```
sas: I think you're duplicating what `kReasonThreadSpawn` should be. We shouldn't be adding this. You'll also see that `kReasonThreadSpawn` exists but isn't used anywhere.

As a side note, if you add something here, you should be adding it to `Stringify::StopReason` as well. You won't have to care about this if you directly use `kReasonThreadSpawn` though.

Sources/Target/Linux/Process.cpp
```diff
@@ -237,7 +272,7 @@ ErrorCode Process::wait() {
       // Killing the main thread?
       //
       // Note(sas): This might be buggy; the main thread exiting
-      // doesn't mean that the process is dying.
+      // doesn't mean that the  process is dying.
```
sas: This looks like an incorrect whitespace change.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,19 +215,54 @@ ErrorCode Process::wait() {
...
-      // the SIGSTOP was queued, and now we're getting this notification. We
-      // have to ignore this and just let the thread continue.
-      // If `stepping` is true, it means the thread was actually being
-      // single-stepped before stopping, so instead of doing a `resume()`, we
-      // have to do a new `step()`.
-      if (stepping) {
+      // If the _stopInfo.event is kEventNone it means the thread is either
+      // 1. stopped because a sibling thread hit a breakpoint and ds2
+      //    sent a SIGSTOP to pause this thread while the breakpoint is
+      //    evaluated. In this case we just call continue.
```
sas: `s/call continue/resume the thread/g`

There are other instances of this below.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,19 +215,54 @@ ErrorCode Process::wait() {
...
     case StopInfo::kEventNone:
-      // If the thread is stopped for no reason, it means the debugger (ds2)
-      // sent a SIGSTOP to it while it was already stopped for another reason,
-      // the SIGSTOP was queued, and now we're getting this notification. We
-      // have to ignore this and just let the thread continue.
-      // If `stepping` is true, it means the thread was actually being
-      // single-stepped before stopping, so instead of doing a `resume()`, we
-      // have to do a new `step()`.
-      if (stepping) {
+      // If the _stopInfo.event is kEventNone it means the thread is either
```
sas: `s/the _stopInfo.event/_stopInfo.event/`

Sources/Target/Linux/Process.cpp
```diff
@@ -215,19 +215,54 @@ ErrorCode Process::wait() {
...
+      // 1. stopped because a sibling thread hit a breakpoint and ds2
```
sas: `s/a sibling thread/another thread of the inferior/`

Sources/Target/Linux/Process.cpp
```diff
@@ -215,19 +215,54 @@ ErrorCode Process::wait() {
...
+      //    sent a SIGSTOP to pause this thread while the breakpoint is
+      //    evaluated. In this case we just call continue.
+      // 2. stopped because this thread called clone(3). In this case
```
sas: `s/clone(3)/clone(2)/`

Sources/Target/Linux/Process.cpp
```diff
@@ -215,19 +215,54 @@ ErrorCode Process::wait() {
...
+      //    we have two subcases:
+      //    a. The thread was running and hit the clone, stopped
+      //       and we just continue it and continue the child as well.
+      //       The loop will continue and the next waitpid will handle the
+      //       freshly cloned thread. In this case we we call continue as well
```
sas: `s/freshly cloned/freshly spawned/`
`s/we we/we/`

Sources/Target/Linux/Process.cpp
```diff
@@ -215,19 +215,54 @@ ErrorCode Process::wait() {
...
+      //       In this case we can run into a race condition with the
+      //       waitpid(3) call and thus have to handle the cloned thread
+      //       and the currentThread at once. So we call step and then
+      //       beforeResume the cloned thread.
+      //
+      // We also have another theoretically possible case where we are not
+      // stopped by a clone but still stepping while encountering a kEventNone.
+      // This should not happen and we just assert that it does not.
+      //
+      if (_currentThread->_stopInfo.reason == StopInfo::kReasonClone &&
```
sas: You can probably re-order this if/else if/else block to avoid having more than one `goto`, and check early for the fatal error you have at line 259.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,19 +215,54 @@ ErrorCode Process::wait() {
...
+            ::waitpid(clonedChildTID, &clonedChildStatus, WNOHANG | __WALL);
+        DS2LOG(Debug, "child thread tid %lu %s", clonedChildTID,
+               Stringify::WaitStatus(clonedChildStatus));
         _currentThread->step();
+        auto newThread = new Thread(this, clonedChildTID);
+
+        newThread->beforeResume();
+        goto continue_waiting;
+      } else if (stepping) {
+        DS2LOG(Fatal, "We should never see a case where we're 1. stopped for "
```
sas: I think you can use `DS2BUG()` and you won't need a `return`.

Headers/DebugServer2/Types.h
lanza: 👍

Sources/Target/Linux/Process.cpp
lanza: I can't count.

Sources/Target/Linux/Process.cpp
lanza: > we we

so french

Sources/Host/POSIX/PTrace.cpp
lanza: I don't think we should have a "high level" method mapping to such a generic low level function. I think it would be better to have multiple methods that do some high level task such as `getClonedTID` that all map to the same `ptrace(PTRACE_GETEVENTMSG,...)` call. Hence the name of this method. This method is only to be used to get the TID of a cloned(should be spawned?) thread.

Sources/Host/POSIX/PTrace.cpp
lanza: Yup, confused the purpose with some subset of flakes with this change.

xiaobai: The whitespace is also strange, looks like some lines are double-indented?

Sources/Host/POSIX/PTrace.cpp
sas: I would agree with you, except for the fact that what you are describing is not possible. `PTRACE_GETEVENTMSG` does different things depending on what the last event that occurred was.

Like in your proposal, all the methods (like `getClonedTID()`) would have the exact same code: a simple call to `ptrace(PTRACE_GETEVENTMSG, ...)`.

lanza:
@sas Please re-review.

xiaobai:
@lanza: You should rebase+update this PR to get the tests to run again. Android-ARM should no longer fail because of your change to `run-lldb-scripts.sh`, meaning we'll get a better signal about this change.

Sources/Host/Linux/PTrace.cpp
```diff
@@ -285,6 +285,16 @@ ErrorCode PTrace::getSigInfo(ProcessThreadId const &ptid, siginfo_t &si) {
   return kSuccess;
 }
 
+ErrorCode PTrace::getSpawnedTID(ProcessId pid, unsigned long &clonedTID) {
```
sas: This has the same issue as what we had before. The function name doesn't correspond to what the function does exactly.

Sources/Target/Linux/Process.cpp
```diff
@@ -212,20 +213,58 @@ ErrorCode Process::wait() {
...
+        spawnedThreadTID =
+            ::waitpid(spawnedThreadTID, &spawnedThreadStatus, WNOHANG | __WALL);
+        DS2LOG(Debug, "child thread tid %lu %s", spawnedThreadTID,
+               Stringify::WaitStatus(spawnedThreadStatus));
         _currentThread->step();
+        auto newThread = new Thread(this, spawnedThreadTID);
+
+        newThread->beforeResume();
+      } else if (stepping) {
+        DS2BUG("We should never see a case where we're\n"
```
sas: Multi-line logs will fuck up the output format. Make it something more concise like
`DS2BUG("Inconsistent thread stop info");`

Sources/Target/Linux/Process.cpp
```diff
@@ -212,20 +213,58 @@ ErrorCode Process::wait() {
 
     stepping = _currentThread->_state == Thread::kStepped;
     _currentThread->updateStopInfo(status);
+    didSpawnThread =
```
sas: This local variable is useless it seems. I'd do the check when you need it, given that you use this result only once. It makes the code easier to read by not having to go back to check where the value comes from.

Support/Testing/Blacklists/ds2/x86_64.blacklist
```diff
@@ -1,2 +0,0 @@
-xfail
```
sas: Does that delete the file or just empty it? The github UI doesn't make it clear.

lanza: The file itself is removed.

Sources/Target/Linux/Process.cpp
lanza: Fixed.

Support/Testing/Blacklists/ds2/x86_64.blacklist
lanza: It's deleted.

Sources/Target/Linux/Process.cpp
lanza: Yea true, I tried cleaning it up and some things failed because of the `goto`s and crossing variable definitions. Forgot to undo it.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
-      // If `stepping` is true, it means the thread was actually being
-      // single-stepped before stopping, so instead of doing a `resume()`, we
-      // have to do a new `step()`.
-      if (stepping) {
+      // If _stopInfo.event is kEventNone it means the thread is either
+      // 1. stopped because another thread of the inferior hit a breakpoint
+      //    and ds2 sent a SIGSTOP to pause this thread while the breakpoint
+      //    is evaluated. In this case we just resume the thread.
+      // 2. stopped because this thread called clone(2) (or similar e.g.
+      //    Pthread_spawn(2)). In this case we have two subcases:
```
xiaobai: What is `Pthread_spawn(2)`? Did you mean something like `pthread_create`?

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+      //       continue_waiting. The loop will continue and the next waitpid
+      //       will handle the freshly spawned thread. In this case we call
+      //       resume as well.
+      //    b. We were single stepping (c.f. the `stepping` var) and
+      //       were stopped by a SIGTRAP for the step while simultaneously
+      //       being stopped via the SIGTRAP for the clone/spawn.
+      //       In this case we can run into a race condition with the
+      //       waitpid(3) call and thus have to handle the cloned thread
+      //       and the currentThread at once. So we call step on the
+      //       _currentThreadand then beforeResume the cloned thread.
```
xiaobai: `_currentThreadand` -> `_currentThread and`

Support/Scripts/docker-utilties.sh
```diff
@@ -0,0 +1,8 @@
+
```
xiaobai: This file probably shouldn't be in this PR/Commit. :P

lanza:
@sas

lanza:
@sas Fixed the commit title.

Sources/Host/Linux/PTrace.cpp
```diff
@@ -285,6 +285,16 @@ ErrorCode PTrace::getSigInfo(ProcessThreadId const &ptid, siginfo_t &si) {
   return kSuccess;
 }
 
+ErrorCode PTrace::getEventMessage(ProcessId pid, unsigned long &data) {
+  auto res = wrapPtrace(PTRACE_GETEVENTMSG, pid, nullptr, &data);
```
sas:
```
if (wrapPtrace(PTRACE_GETEVENTMSG, pid, nullptr, &data) < 0) {
    return Platform::TranslateError();
}
```

Headers/DebugServer2/Host/Linux/PTrace.h
```diff
@@ -66,6 +66,7 @@ class PTrace : public POSIX::PTrace {
 
 public:
   ErrorCode getSigInfo(ProcessThreadId const &ptid, siginfo_t &si) override;
+  ErrorCode getEventMessage(ProcessId pid, unsigned long &data);
```
sas: I think the type of `pid` should be `ProcessThreadId` and `getEventMessage()` should convert it to a `pid_t` using `CHK(ptidToPid(ptid, pid));` (see the implementation of `PTrace::getSigInfo()`).

Sources/Target/Linux/Process.cpp
```diff
@@ -175,9 +175,9 @@ ErrorCode Process::checkMemoryErrorCode(uint64_t address) {
 
 ErrorCode Process::wait() {
   int status, signal;
-  bool stepping;
   ProcessInfo info;
   ThreadId tid;
+  bool stepping;
```
sas: Why move this?

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+      //       being stopped via the SIGTRAP for the clone/spawn.
+      //       In this case we can run into a race condition with the
+      //       waitpid(3) call and thus have to handle the cloned thread
+      //       and the currentThread at once. So we call step on the
+      //       _currentThread and then beforeResume the cloned thread.
+      //
+      // We also have another theoretically possible case where we are not
+      // stopped by a clone but still stepping while encountering a kEventNone.
+      // This should not happen and we just assert that it does not.
+      //
```
sas: Empty comment line.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
-      // If the thread is stopped for no reason, it means the debugger (ds2)
-      // sent a SIGSTOP to it while it was already stopped for another reason,
-      // the SIGSTOP was queued, and now we're getting this notification. We
-      // have to ignore this and just let the thread continue.
-      // If `stepping` is true, it means the thread was actually being
-      // single-stepped before stopping, so instead of doing a `resume()`, we
-      // have to do a new `step()`.
-      if (stepping) {
+      // If _stopInfo.event is kEventNone it means the thread is either
+      // 1. stopped because another thread of the inferior hit a breakpoint
```
sas: Technically, it's because another thread of the inferior stopped for any reason, received a signal, etc.

I'd probably word this as "another thread of the inferior stopped (e.g.: because of a breakpoint, or received a signal, etc)"/

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+      //    and ds2 sent a SIGSTOP to pause this thread while the breakpoint
+      //    is evaluated. In this case we just resume the thread.
+      // 2. stopped because this thread called clone(2) (or similar e.g.
+      //    pthread_create). In this case we have two subcases:
+      //    a. The thread was running and hit the clone/spawn and thus was
+      //       send a SIGTRAP and stopped. We just resume it and goto
```
sas: `s/was send a SIGTRAP and stopped/stopped with a SIGTRAP/` (you have a typo with send, and technically, nobody sends anything, the kernel just stops the thread and reports the status as if it were stopped with a SIGTRAP.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+      // stopped by a clone but still stepping while encountering a kEventNone.
+      // This should not happen and we just assert that it does not.
+      //
+
+      if (_currentThread->_stopInfo.reason == StopInfo::kReasonThreadSpawn &&
+          stepping) { //(2b)
+        unsigned long spawnedThreadTID = 0;
+        // ptrace(PTRACE_GETEVENTMSG, pid, addr, data) sets the data argument
+        // to the spawned thread's tid
+        auto error = ptrace().getEventMessage(_pid, spawnedThreadTID);
```
sas: You should avoid using `auto` when the return type is not super obvious. In this specific case we use the `CHK()` macro for trivial error handling like this.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+      //
+      // We also have another theoretically possible case where we are not
+      // stopped by a clone but still stepping while encountering a kEventNone.
+      // This should not happen and we just assert that it does not.
+      //
+
+      if (_currentThread->_stopInfo.reason == StopInfo::kReasonThreadSpawn &&
+          stepping) { //(2b)
+        unsigned long spawnedThreadTID = 0;
+        // ptrace(PTRACE_GETEVENTMSG, pid, addr, data) sets the data argument
```
sas: Useless comment :p

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+      //       _currentThread and then beforeResume the cloned thread.
+      //
+      // We also have another theoretically possible case where we are not
+      // stopped by a clone but still stepping while encountering a kEventNone.
+      // This should not happen and we just assert that it does not.
+      //
+
+      if (_currentThread->_stopInfo.reason == StopInfo::kReasonThreadSpawn &&
+          stepping) { //(2b)
+        unsigned long spawnedThreadTID = 0;
```
sas: (This is nitpicking.)

No need to initialize the value of `spawnedThreadTID` if you're gonna set it with the next ptrace call. Also, "ThreadTID" is redundant; `spawnedThreadID` would be good.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+        // ptrace(PTRACE_GETEVENTMSG, pid, addr, data) sets the data argument
+        // to the spawned thread's tid
+        auto error = ptrace().getEventMessage(_pid, spawnedThreadTID);
+        if (error != 0) {
+          return error;
+        }
+
+        int spawnedThreadStatus = 0;
+        spawnedThreadTID =
```
sas: I don't think you should reuse `spawnedThreadTID` for error checking after having set it to the new thread's ID.
Also, you should actually do the error checking. Here, it seems like you're assigning the result to a variable and never checking the contents.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+          stepping) { //(2b)
+        unsigned long spawnedThreadTID = 0;
+        // ptrace(PTRACE_GETEVENTMSG, pid, addr, data) sets the data argument
+        // to the spawned thread's tid
+        auto error = ptrace().getEventMessage(_pid, spawnedThreadTID);
+        if (error != 0) {
+          return error;
+        }
+
+        int spawnedThreadStatus = 0;
```
sas: Useless initialization.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+        spawnedThreadTID =
+            ::waitpid(spawnedThreadTID, &spawnedThreadStatus, WNOHANG | __WALL);
+        DS2LOG(Debug, "child thread tid %lu %s", spawnedThreadTID,
+               Stringify::WaitStatus(spawnedThreadStatus));
         _currentThread->step();
+        auto newThread = new Thread(this, spawnedThreadTID);
```
sas: Using `auto` here is fine for example.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+
```
sas: Useless blank line.

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,57 @@ ErrorCode Process::wait() {
...
+        newThread->beforeResume();
+      } else if (stepping) {
+        // We should never see a case where we're:
+        //   1. stopped for event kEventNone
+        //   2. stepping
+        //   3. not stopped for reason kReasonThreadSpawn
+        DS2BUG("Inconsistent thread stop info.");
```
sas: No comma and no capital letter for log messages.

lanza: Period?

xiaobai:
@sas please re-review. :D

Sources/Target/Linux/Process.cpp
```diff
@@ -215,17 +215,56 @@ ErrorCode Process::wait() {
...
+      //       and the currentThread at once. So we call step on the
+      //       _currentThread and then beforeResume the cloned thread.
+      //
+      // We also have another theoretically possible case where we are not
+      // stopped by a clone but still stepping while encountering a kEventNone.
+      // This should not happen and we just assert that it does not.
+
+      if (_currentThread->_stopInfo.reason == StopInfo::kReasonThreadSpawn &&
+          stepping) { //(2b)
+        ProcessThreadId ptid(_pid);
```
sas: Unused local variable.

PR #632, opened 2018-06-01
Add support in run-lldb-tests.sh to exclude centos tests

lanza:
`FAIL: TestExprs2.ExprCommands2TestCase.test_more_expr_commands_dwarf (expression_command/test/TestExprs2.py)`

consistently occurs on master. Add this to follow up with a centos.blacklists addition to hide this flake.


sas:
Ok so I'm gonna reject this PR until it includes an associated blacklist then.

lanza:
ce bon

PR #634, opened 2018-06-04
Print an error when wget is not found in prepare-android-ndk.sh

sas:


sas:
Honestly I was going to change the script to use curl but we don't really have a guarantee it exists on any Linux installation so I opted to do this instead.

PR #635, opened 2018-06-04
Make Process::ptrace() return a more specific type

sas:
This will make it easier to add Linux-only functionality to PTrace for
instance.

### DIFF
--- a/Headers/DebugServer2/Target/Linux/Process.h
+++ b/Headers/DebugServer2/Target/Linux/Process.h
@@ -53,7 +53,7 @@ public:
   ErrorCode wait() override;
 
 public:
-  Host::POSIX::PTrace &ptrace() const override;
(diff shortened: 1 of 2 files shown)


sas:
Looks like you already tried this in PR#627 but I'm not sure why you abandoned it. I simply tried to build my PR, not run tests with it, so not sure what issues you saw.

sas:
> I remember you telling me that this also broke some functionality elsewhere for some reason. Is that no longer the case?

I have really no idea. If I said that I can't remember why. This is the kind of change you can usually validate by just building the source, and I was able to successfully build so...

lanza:
I think I confused the flakes I disabled in the centos PR as ones being caused by this change. That or ones that are caused by my VM being too slow for Linux runs.

PR #639, opened 2018-06-06
Change run-lldb-tests.sh to use 127.0.0.1 instead of localhost or *

lanza:
Android-ARM is failing to create a socket given localhost or * as the
address. Darwin is definitively failing on `::getaddrinfo` when fed *.
So, for now:

Change run-lldb-tests.sh to use 127.0.0.1 instead of localhost or *
as the argument for the address for the socket connection


xiaobai:
I'm excited to see what will happen.

Nonetheless, I wish I could post macros :(

lanza:
You need Memojis. Time to upgrade.

xiaobai:
not sure ds2 can handle all this green.

PR #641, opened 2018-06-08

lanza:


lanza:
Please confer https://github.com/facebook/ds2/pull/523 for old comments etc.

xiaobai:
It looks like Android-ARM might actually be failing because of this change. :/

xiaobai:
I was good with this when fjricci did this, and I looked at your local patch and it looks fine. I did notice that with your Lifetime changes, you now have a conflict, so you should update that.

I would be more comfortable with @sas reviewing this since he never reviewed the original patchset.

xiaobai:
ping @sas

lanza:
Pinging @sas

PR #642, opened 2018-06-09
Add typesafe bitwise enums and convert Lifetime to use it

lanza:
Previously an enum life `Lifetime` was declared as a trio of `unsigned int`s with some support from the compiler to suggest that you don't use them incorrectly. However, C++11 introduced `enum class`es various template techniques to make this more typesafe.

1. Convert `BreakpointManager::Lifetime` from a bare `enum` to an `enum class`.
2. Template instantiate the necessary bitwise operators for the bitwise mathematics done in `BreakpointManager` (Headers/DebugServer2/Core/BreakpointManager.h:115).
3. Convert the usage sites of `Lifetime` to the new format.

### DIFF
--- a/Headers/DebugServer2/Core/BreakpointManager.h
+++ b/Headers/DebugServer2/Core/BreakpointManager.h
@@ -11,6 +11,7 @@
 #pragma once
 
 #include "DebugServer2/Target/ProcessDecl.h"
+#include "DebugServer2/Utils/Enums.h"
 
 #include <functional>
 
@@ -18,10 +19,11 @@ namespace ds2 {
 
 class BreakpointManager {
 public:
-  enum Lifetime : unsigned int {
-    kLifetimePermanent = (1 << 0),
-    kLifetimeTemporaryOneShot = (1 << 1),
-    kLifetimeTemporaryUntilHit = (1 << 2),
+  enum class Lifetime : unsigned int {
+    None = 0,
+    Permanent = (1 << 0),
+    TemporaryOneShot = (1 << 1),
+    TemporaryUntilHit = (1 << 2),
   };
 
   enum Mode {
@@ -109,3 +111,4 @@ public:
   virtual bool fillStopInfo(Target::Thread *thread, StopInfo &stopInfo) = 0;
 };
 } // namespace ds2
+ENABLE_BITMASK_OPERATORS(ds2::BreakpointManager::Lifetime);
--- a/Headers/DebugServer2/Utils/Enums.h
+++ b/Headers/DebugServer2/Utils/Enums.h
@@ -1,0 +1,48 @@
+//
+// Copyright (c) 2014-present, Facebook, Inc.
+// All rights reserved.
+//
+// This source code is licensed under the University of Illinois/NCSA Open
+// Source License found in the LICENSE file in the root directory of this
+// source tree. An additional grant of patent rights can be found in the
+// PATENTS file in the same directory.
+//
+
+template <typename Enum> struct EnableBitMaskOperators {
+  static const bool enable = false;
+};
+
+template <typename Enum>
+typename std::enable_if<EnableBitMaskOperators<Enum>::enable, Enum>::type
+operator|(Enum lhs, Enum rhs) {
+  using underlying = typename std::underlying_type<Enum>::type;
+  return static_cast<Enum>(static_cast<underlying>(lhs) |
+                           static_cast<underlying>(rhs));
+}
+
+template <typename Enum>
+typename std::enable_if<EnableBitMaskOperators<Enum>::enable, Enum>::type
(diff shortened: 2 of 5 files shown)


Headers/DebugServer2/Core/BreakpointManager.h
```diff
@@ -109,3 +111,5 @@ class BreakpointManager {
   virtual bool fillStopInfo(Target::Thread *thread, StopInfo &stopInfo) = 0;
 };
 } // namespace ds2
+
+ENABLE_BITMASK_OPERATORS(ds2::BreakpointManager::Lifetime);
```
sas: I'd put this right after line 26, so everything regarding the `Lifetime` type will be grouped together.

Headers/DebugServer2/Utils/Enums.h
```diff
@@ -0,0 +1,48 @@
...
+// Copyright (c) 2014-present, Facebook, Inc.
+// All rights reserved.
+//
+// This source code is licensed under the University of Illinois/NCSA Open
+// Source License found in the LICENSE file in the root directory of this
+// source tree. An additional grant of patent rights can be found in the
+// PATENTS file in the same directory.
+//
+
+template <typename Enum> struct EnableBitMaskOperators {
```
sas: You should wrap all if this in 
```
namespace ds2 {
namespace Utils {
}
}
```

Sources/Core/BreakpointManager.cpp
```diff
@@ -143,10 +142,10 @@ void BreakpointManager::disable(Target::Thread *thread) {
   //
   auto it = _sites.begin();
   while (it != _sites.end()) {
-    it->second.lifetime =
-        static_cast<Lifetime>(it->second.lifetime & ~kLifetimeTemporaryOneShot);
-    if (!it->second.lifetime) {
-      // refs should always be 0 unless we have a kLifetimePermanent breakpoint.
+    it->second.lifetime = it->second.lifetime & ~Lifetime::TemporaryOneShot;
+    if (it->second.lifetime == Lifetime::None) {
```
sas: Nice.

Headers/DebugServer2/Utils/Enums.h
```diff
@@ -0,0 +1,48 @@
...
+  static const bool enable = false;
+};
+
+template <typename Enum>
+typename std::enable_if<EnableBitMaskOperators<Enum>::enable, Enum>::type
+operator|(Enum lhs, Enum rhs) {
+  using underlying = typename std::underlying_type<Enum>::type;
```
sas: `s/underlying/Underlying/`

Headers/DebugServer2/Utils/Enums.h
```diff
@@ -0,0 +1,48 @@
...
+  return static_cast<Enum>(static_cast<underlying>(lhs) |
+                           static_cast<underlying>(rhs));
+}
+
+template <typename Enum>
+typename std::enable_if<EnableBitMaskOperators<Enum>::enable, Enum>::type
```
sas: You should also mark these `constexpr` so you can use them in a context where constant expressions are required (e.g.: in a `switch` statement).

Headers/DebugServer2/Utils/Enums.h
xiaobai: Additionally you're `using Underlying = typename std::underlying_type<Enum>::type;` in basically every one of these. Is there anyway you can pull it out and write it once? :P

Headers/DebugServer2/Core/BreakpointManager.h
lanza: The namespacing goes awry. The individual operator definitions aren't told to look into any particular namespace and so the definition of the individual `EnableBitMaskOperators` struct needs to match be defined in terms of the fully qualified typename e.g. `ds2::Something::Other::Thing`.

At least that's what I've gathered from the completely opaque error messages.

Headers/DebugServer2/Utils/Enums.h
lanza: AFIAK you can't. Similar issue as above. You'd run into problems with the operators being defined in a different namespace than the type they are defined for. `ds2::utils::operator|` generic over `ds::BreakpointManager::Lifetime`.

Headers/DebugServer2/Utils/Enums.h
lanza: Indeed. You could clearly wrap this in more macros and what not but I'm not sure layering this in more levels of indirection makes it any clearer or better. I debated on a `unaryop` and `binaryop` template of some sort but meh.

lanza:
@sas So this doesn't work with older versions of gcc. (e.g. the 5.4 on the docker image).

PR #643, opened 2018-06-12
Remove hacky testsuite workaround on android

xiaobai:
I removed the hacky patch to see what would happen. Surprisingly, it passed locally. Let's see what CircleCI thinks.


xiaobai:
The world is not ready for this.

PR #646, opened 2018-07-17
Add arm64 to CircleCI Dockerfile

lanza:
Add arm64 to the CircleCI Dockerfile to enable aarch64 testing


lanza:
In progress, there are a handful of other things to do. This is just one.

PR #648, opened 2018-07-18
Add error messages for some run-lldb-tests crashes

lanza:
`set -eu` causes `which` to exit the program when the argument does not
exist. This causes occasional crashes in run-lldb-tests.sh where no
error is reported.

Add corresponding error messages to the crash conditions.


sas:
This is a bit meh. You should do a function for these instead of copying the same logic everywhere. I think there's a file for helper functions like `common.sh` or something.

PR #649, opened 2018-07-18
Add flags to support scripts

mattmacumber:
Add flags to support scripts to allow flag-based arguments in addition to environment-based arguments.


lanza:
Looks good to me. Thanks, Matt!

PR #650, opened 2018-07-19
Fix two bugs involving the timing of waitpid events

lanza:
Due to the non-deterministic nature of the signals being sent to the
cloned thread and the parent thread, `waitpid(... WNOHANG)` could end up
returning empty due to not having any child to wait for.

Remove `WNOHANG` so that we will wait for a child thread to be stopped
on clone given that we were just ignored by geteventmsg that there is a
thread to wait for.

### DIFF
--- a/Sources/Target/Linux/Process.cpp
+++ b/Sources/Target/Linux/Process.cpp
@@ -245,17 +245,29 @@ ErrorCode Process::wait() {
         CHK(ptrace().getEventMessage(_pid, spawnedThreadIdData));
         auto const spawnedThreadId = static_cast<ThreadId>(spawnedThreadIdData);
 
-        int spawnedThreadStatus;
-        ThreadId const returnedThreadId =
-            ::waitpid(spawnedThreadId, &spawnedThreadStatus, WNOHANG | __WALL);
-        if (spawnedThreadId != returnedThreadId)
-          return kErrorProcessNotFound;
-        DS2LOG(Debug, "child thread tid %d %s", returnedThreadId,
-               Stringify::WaitStatus(spawnedThreadStatus));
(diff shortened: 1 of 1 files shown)


lanza:
Observed it. It's decently reproducible locally, too. I think I missed it up until this point because I've always tested with `--log` on run-lldb-tests.py and the log calls give enough time for the kernel to send the signal before `waitpid` is called. That or some of the refactoring done over the last few revisions introduced this bug. I wish GitHub had revisions like Phabricator...

The stop info being `StopInfo::kReasonThreadSpawn` guarantees that there is a child thread pending to be waited on. So a hanging `waitpid` should work here. I'm trying to think of a time where this wouldn't be the case but 100% tests passing makes it seem okay.

Sources/Target/Linux/Process.cpp
```diff
@@ -245,17 +245,29 @@ ErrorCode Process::wait() {
...
+                 Stringify::WaitStatus(spawnedThreadStatus));
+
+          _currentThread->step();
+          auto newThread = new Thread(this, returnedThreadId);
+          newThread->beforeResume();
+        } else {
+          // The new thread corresponding to this kReasonThreadSpawn was already
+          // caught by waitpid(2). Therefore, we only need to step the current
+          // thread.
+          _currentThread->step();
```
lanza: I think that this `_currentThread->step()` could be brought out of this `else` statement along with the one above. But I find that the explanation of the logic is much clearer just repeating the `_currentThread->step()`. And given how subtle these potential race conditions are that we're working with here I figured clarity is the best option.

lanza:
@sas Ready for review.

PR #651, opened 2018-07-19
Add some facilities to debug tests and clean up run-lldb-tests.py

lanza:
Clean up some repetitive code, clean up some formatting and add an
option to run the test suite under control of PUDB.

### DIFF
--- a/Support/Scripts/common.sh
+++ b/Support/Scripts/common.sh
@@ -85,3 +85,8 @@ get_android_ndk_dir() {
 check_program_exists() {
   which "$1" >/dev/null 2>&1
 }
+
+get_program_path() {
+  local program=$(which $1) || die "$1 not found"
+  echo "${program}"
+}
--- a/Support/Scripts/run-lldb-tests.sh
+++ b/Support/Scripts/run-lldb-tests.sh
@@ -62,6 +62,7 @@ opt_log=false
 opt_pudb=false
 opt_strace=false
 opt_use_lldb_server=false
+opt_dotest_extra_args=""
 
 while test $# -gt 0; do
   case "$1" in
@@ -78,6 +79,7 @@ while test $# -gt 0; do
 	    TARGET="$1";;
     --platform) shift
 	    PLATFORM="$1";;
+    --dotest-extra-args) shift; opt_dotest_extra_args="$1";;
     *) die "Unknown option \`$1'.";;
   esac
   shift
@@ -287,5 +289,7 @@ export LLDB_TEST_TIMEOUT=8m
 if [ "${LLDB_TESTS-all}" != "all" ]; then
   args+=(-p "$LLDB_TESTS")
 fi
+
+args+=(${opt_dotest_extra_args})
 
 "${python_exe[@]}" dotest.py "${args[@]}"
--- a/Support/Scripts/run-lldb-tests.sh
+++ b/Support/Scripts/run-lldb-tests.sh
@@ -73,12 +73,9 @@ while test $# -gt 0; do
     --pudb) opt_pudb=true;;
     --strace) opt_strace=true;;
     --use-lldb-server) opt_use_lldb_server=true;;
-    --lldb-tests) shift
-	    LLDB_TESTS="$1";;
-    --target) shift
-	    TARGET="$1";;
-    --platform) shift
-	    PLATFORM="$1";;
+    --platform) PLATFORM=1;;
+    --lldb-tests) shift; LLDB_TESTS="$1";;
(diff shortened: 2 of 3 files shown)


lanza:
I added the PUDB stuff because I regularly end up stepping into dotest.py in the debugger. The lldb testsuite uses the `inspect` module whose functionality breaks when using a debugger. So I have to add a patch on top before adding the PUDB. I've been maintaining this on the side for some time but it's generically useful and clumsy to cherry-pick around.

Support/Scripts/common.sh
```diff
@@ -85,3 +85,8 @@ get_android_ndk_dir() {
 check_program_exists() {
   which "$1" >/dev/null 2>&1
 }
+
+get_program() {
+  program=$(which $1) || die "$1 not found"
```
sas: local program=“$(which “$1”)” || die ...

Support/Scripts/common.sh
```diff
@@ -85,3 +85,8 @@ get_android_ndk_dir() {
 check_program_exists() {
   which "$1" >/dev/null 2>&1
 }
+
+get_program() {
```
sas: Nit: get_program_path

Support/Scripts/run-lldb-tests.sh
```diff
@@ -55,6 +60,7 @@ opt_no_upstream_blacklists=false
 opt_log=false
 opt_strace=false
 opt_use_lldb_server=false
+DOSTEP_EXTRA_ARGS=""
```
sas: You should probably name this opt_dostep_... to match with the other option variables.

Support/Scripts/run-lldb-tests.sh
```diff
@@ -65,17 +71,19 @@ while test $# -gt 0; do
     --strace) opt_strace=true;;
     --use-lldb-server) opt_use_lldb_server=true;;
     --lldb-tests) shift
```
sas: The formatting of these lines is really odd.

Support/Scripts/run-lldb-tests.sh
```diff
@@ -65,17 +71,19 @@ while test $# -gt 0; do
...
+    --pudb) PUDB="-m pudb";;
+    --dostep-extra-args) shift
+      DOSTEP_EXTRA_ARGS="$1";;
     *) die "Unknown option \`$1'.";;
   esac
   shift
 done
 
-TARGET="${TARGET-${CIRCLE_JOB}}"
+TARGET="${TARGET-${CIRCLE_JOB-}}"
```
sas: Don’t we need to have at least one of these two set?

Support/Scripts/run-lldb-tests.sh
```diff
@@ -65,17 +71,19 @@ while test $# -gt 0; do
...
     --lldb-tests) shift
-	    LLDB_TESTS="$1";;
+      LLDB_TESTS="$1";;
     --target) shift
-	    TARGET="$1";;
-    --platform) shift
-	    PLATFORM="$1";;
+      TARGET="$1";;
+    --platform) PLATFORM=1;;
+    --pudb) PUDB="-m pudb";;
```
sas: Instead of storing this as a list of flags, I’d probably have opt_pudb=true, and check the conditional later when building the “args” array.

sas:
Also, this is a combination of multiple changes. Can you separate them out in individual commits?

Can we upstream the local lldb diff that you’re adding?

Support/Scripts/run-lldb-tests.sh
lanza: It's not an option, though. `arg_dostep_extra_args`?

Support/Scripts/run-lldb-tests.sh
lanza: No, dostep.py defaults to the value derived from the python `sys.platform` or similar for the host given an unset `TARGET`.

Support/Scripts/run-lldb-tests.sh
sas: It's not true/false but it's a command-line option. No?

Support/Scripts/run-lldb-tests.sh
sas: Nice. If that works I'm ok with this then.

Support/Scripts/run-lldb-tests.sh
lanza: It can't be to the args array, it has to be called before the python file name. `python -m module args-for-the-module`.

PR #653, opened 2018-07-25
Rewrite start-android-emulator in python3.5

lanza:


.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ check_dirty_step: &check_dirty_step
 start_android_emulator_step: &start_android_emulator_step
   run:
     name: Start Android emulator
-    command: ./Support/Scripts/start-android-emulator.sh
+    command: ./Support/Scripts/start-android-emulator
```
sas: Just to be consistent with the rest, you should name it `start-android-emulator.py`.

Support/Scripts/start-android-emulator
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
sas: 👍

Support/Scripts/start-android-emulator
```diff
@@ -0,0 +1,87 @@
...
+
+import os
+import argparse
+import subprocess
+from typing import Dict, Any, Tuple
+import sys
+import copy
+
+
+def run_argument_parser() -> argparse.Namespace:
```
sas: Fancy typing. I like it.

Support/Scripts/start-android-emulator
```diff
@@ -0,0 +1,87 @@
...
+            args.taget = os.environ.get("TARGET", None)
+        elif not os.environ.get("CIRCLE_JOB", None):
+            args.target = os.environ.get("CIRCLE_JOB", None)
+        else:
+            print(
+                "either the TARGET env var, the CIRCLE_JOB env var or the --target flag is required"
+            )
+            exit(1)
+
+    if not args.download_path:
```
sas: Instead of doing this you can specify a default value in argparse.

Support/Scripts/start-android-emulator
```diff
@@ -0,0 +1,87 @@
...
+    parser.add_argument("--target", help="the target emulator to run")
+    parser.add_argument(
+        "--download-path", help="the directory where the ndk and sdk are installed"
+    )
+
+    args = parser.parse_args()
+    return args
+
+
+def clean_arguments(args: argparse.Namespace) -> argparse.Namespace:
```
sas: I think this is messy because of the stupid stuff we're doing with environment variables. We could just get rid of `$TARGET` because I think we don't really use it, and simply have one spot in `run_argument_parser()` to read `$CIRCLE_JOB` and fill in the info.

Support/Scripts/start-android-emulator
sas: Nit: The name is verbose. I'd do something like `parse_args()` or `parse_arguments()`.

Support/Scripts/start-android-emulator
```diff
@@ -0,0 +1,87 @@
...
+def get_arch_from_target(target: str) -> Tuple[str, str]:
+    target_arch_pairs = {
+        "Android-ARM": ("emulator64-arm", "arm"),
+        "Android-ARM64": ("emulator", "arm64"),
+        "Android-x86": ("emulator64-x86", "x86"),
+    }
+    archs = target_arch_pairs.get(target, None)
+    if archs == None:
+        print("either the TARGET env var or the --target flag is required")
+        exit(1)
```
sas: Did you test this? Shouldn't this be `sys.exit(1)` instead?

Support/Scripts/start-android-emulator
```diff
@@ -0,0 +1,87 @@
...
+    else:
+        if not os.environ.get("TARGET", None) == None:
+            args.taget = os.environ.get("TARGET", None)
+        elif not os.environ.get("CIRCLE_JOB", None):
+            args.target = os.environ.get("CIRCLE_JOB", None)
+        else:
+            print(
+                "either the TARGET env var, the CIRCLE_JOB env var or the --target flag is required"
+            )
+            exit(1)
```
sas: Same here, shouldn't this be `sys.exit(1)`?

Support/Scripts/start-android-emulator
```diff
@@ -0,0 +1,87 @@
...
+
+
+def get_arch_from_target(target: str) -> Tuple[str, str]:
+    target_arch_pairs = {
+        "Android-ARM": ("emulator64-arm", "arm"),
+        "Android-ARM64": ("emulator", "arm64"),
+        "Android-x86": ("emulator64-x86", "x86"),
+    }
+    archs = target_arch_pairs.get(target, None)
+    if archs == None:
```
sas: Nit: `if not archs:`

Support/Scripts/start-android-emulator.py
```diff
@@ -0,0 +1,90 @@
...
+    qt_lib_path = "{}/emulator/lib64/qt/lib".format(sdk_dir)
+
+    llp = os.environ.get("LD_LIBRARY_PATH")
+    if llp is None:
+        llp = qt_lib_path
+    else:
+        llp = "{}:{}".format(qt_lib_path, llp)
+    os.environ["LD_LIBRARY_PATH"] = llp
+
+    subprocess.call(
```
xiaobai: Python 3.5 introduces `subprocess.run()`, which I think you should use unless you need this to work on older versions of python.

Support/Scripts/start-android-emulator.py
```diff
@@ -0,0 +1,90 @@
...
+    sdk_dir = "{}/android-sdk-{}".format(args.download_path, sys.platform)
+    emulator = "{}/emulator/{}".format(sdk_dir, emulator_arch)
+    qt_lib_path = "{}/emulator/lib64/qt/lib".format(sdk_dir)
+
+    llp = os.environ.get("LD_LIBRARY_PATH")
+    if llp is None:
+        llp = qt_lib_path
+    else:
+        llp = "{}:{}".format(qt_lib_path, llp)
+    os.environ["LD_LIBRARY_PATH"] = llp
```
xiaobai: Instead of modifying the environment, you should create a clone and modify `LD_LIBRARY_PATH`. Then you can pass the cloned version to `subprocess.run()` which I suggested you use below (or `subprocess.call()` if you don't want to change it).

PR #656, opened 2018-07-31
wip - aarch64 android

lanza:


sas:
Android-ARM64 still seems to fail.

PR #659, opened 2018-08-01
Add proxy arguments to install-android-emulator for centos

lanza:


Support/Scripts/install-android-emulator.sh
```diff
@@ -27,7 +27,11 @@ if [ ! -f "$sdkmanager" ]; then
   rm -f "/tmp/${package_name}"
 fi
 
-echo "y" | "${sdkmanager}" --update
+if [[ $(hostname) =~ facebook ]]; then
+  sdkmanager="${sdkmanager} --verbose --no_https --proxy=http --proxy_host=fwdproxy --proxy_port=8080"
```
sas: If you're gonna do stuff like that you should use a vector instead.

PR #662, opened 2018-08-06
Rewrite start-android-emulator in python3.5

lanza:


.circleci/config.yml
```diff
@@ -136,6 +136,7 @@ jobs:
     <<: *build_and_test_android
     environment:
       PLATFORM: 1
+      TARGET: Android-ARM
```
sas: How did this work before?

xiaobai: Because the scripts used CIRCLE_JOB which is Android-ARM. This line isn't really needed unless you recently modified behavior. :P

sas: Ha. Now it makes sense. Thanks for reminding me of how this works @xiaobai

PR #671, opened 2018-08-18

lanza:

### DIFF
--- a/Sources/Target/Common/ProcessBase.cpp
+++ b/Sources/Target/Common/ProcessBase.cpp
@@ -226,6 +226,11 @@ ErrorCode ProcessBase::readMemoryBuffer(Address const &address, size_t length,
     return error;
   }
 
+  SoftwareBreakpointManager *bpm = softwareBreakpointManager();
+  if (bpm) {
+    bpm->insertStashedInsns(address, length, buffer);
+  }
+
   buffer.resize(nread);
   return kSuccess;
 }
--- a/Sources/Target/Common/ProcessBase.cpp
+++ b/Sources/Target/Common/ProcessBase.cpp
@@ -291,14 +291,6 @@ ErrorCode ProcessBase::beforeResume() {
   if (!isAlive())
     return kErrorProcessNotFound;
 
-  //
-  // Enable software breakpoints.
-  //
-  SoftwareBreakpointManager *bpm = softwareBreakpointManager();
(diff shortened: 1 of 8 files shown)


Sources/Target/Common/ProcessBase.cpp
```diff
@@ -289,8 +289,8 @@ ErrorCode ProcessBase::beforeResume() {
   //
   // Enable software breakpoints.
   //
-  BreakpointManager *bpm = softwareBreakpointManager();
-  if (bpm != nullptr) {
+  SoftwareBreakpointManager *bpm = softwareBreakpointManager();
+  if (bpm) {
```
sas: This should stay as `if (bpm != nullptr)`.

Sources/Target/Common/ProcessBase.cpp
```diff
@@ -305,20 +305,15 @@ ErrorCode ProcessBase::afterResume() {
   }
 
   // Disable breakpoints and try to hit software breakpoints.
-  BreakpointManager *swBpm = softwareBreakpointManager();
-  if (swBpm != nullptr) {
+  SoftwareBreakpointManager *bpm = softwareBreakpointManager();
+  if (bpm) {
```
sas: Same here.

Sources/Target/Common/ProcessBase.cpp
```diff
@@ -344,7 +339,7 @@ HardwareBreakpointManager *ProcessBase::hardwareBreakpointManager() const {
 
 void ProcessBase::prepareForDetach() {
   SoftwareBreakpointManager *bpm = softwareBreakpointManager();
-  if (bpm != nullptr) {
+  if (bpm) {
```
sas: And here.

Sources/Target/Common/ThreadBase.cpp
```diff
@@ -32,9 +32,9 @@ ErrorCode ThreadBase::modifyRegisters(
 }
 
 ErrorCode ThreadBase::beforeResume() {
-  BreakpointManager *bpm = _process->hardwareBreakpointManager();
-  if (bpm != nullptr) {
-    bpm->enable((Target::Thread *)this);
+  HardwareBreakpointManager *bpm = _process->hardwareBreakpointManager();
+  if (bpm) {
```
sas: And here.

PR #674, opened 2018-08-20
Lower cmake required version to 3.9

sas:
I don't think we actually need 3.10, and some build hosts we have
internally use cmake 3.9.


lanza:
If it passes 👍🏼

PR #675, opened 2018-08-24
Add --rerun-all-issues to dotest invocations

lanza:


Support/Scripts/run-lldb-tests.sh
```diff
@@ -292,6 +292,8 @@ if [ "${LLDB_TESTS-all}" != "all" ]; then
   args+=(-p "$LLDB_TESTS")
 fi
 
+args+=('--rerun-all-issues')
```
sas: You should add a comment here to explain why we're doing this.

PR #676, opened 2018-09-02
[test] cxx 17

lanza:
just curious


lanza:
Yea I wasn't planning on landing. Just curious about what would need updated to get 17 in.

PR #677, opened 2018-09-02
[wip] llvm 7.0

lanza:


sas:
I think you're gonna have to update the circleCI docker image to include lldb-7.0 as well. Good stuff though :)

.circleci/config.yml
```diff
@@ -71,12 +71,12 @@ jobs:
...
       - run:
           name: Apply clang-format
-          command: find Sources Headers -type f -exec clang-format-6.0 -i -style=file {} \;
+          command: find Sources Headers -type f -exec clang-format -i -style=file {} \;
       - *check_dirty_step
   Registers:
     <<: *docker_config
     environment:
-      CLANG_FORMAT: clang-format-6.0
+      CLANG_FORMAT: clang-format
```
sas: I think you can delete this env variable if your clang-format is named `clang-format`.

Support/Scripts/run-lldb-tests.sh
```diff
@@ -19,11 +19,11 @@ cherry_pick_patches() {
   # Disabled and enabled tests
   local testingPath="$top/Support/Testing"
 
-  if [[ "$platform_name" = "android" ]]; then
+  #if [[ "$platform_name" = "android" ]]; then
```
sas: You should delete this block if it's not needed anymore instead of just commenting it.

Support/Scripts/run-lldb-tests.sh
```diff
@@ -187,7 +187,7 @@ if [ -n "${TARGET-}" ]; then
   if [[ "${TARGET}" == "Linux-X86_64" || "${TARGET}" == "Linux-X86_64-Clang" ]]; then
     args+=("--arch=x86_64")
     if [[ "${PLATFORM-}" = "1" ]]; then
-      args+=("--excluded" "$blacklist_dir/ds2/llvm_60.blacklist")
+      args+=("--excluded" "$blacklist_dir/ds2/llvm.blacklist")
```
sas: I don't think we should remove the suffix. These are related to new versions of llvm and we should still mark them separately to make resolution easier in the future.

Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,9 +40,38 @@ MAINTAINER Stephane Sezer <sas@fb.com>
 RUN apt-get update
 RUN apt-get install -y software-properties-common python-software-properties wget
 
+# Incase lldb needs built
```
sas: The "utilities" should be in a separate PR and should be clearly marked as optional stuff. Did you check to see if it was possible to just have a command-line flag and enable these when developers need it only?

Also, Incase -> In case.

Support/Testing/CircleCI/Dockerfile
```diff
@@ -51,7 +80,8 @@ RUN apt-get install -y flex bison
 
 # Install x86 compilers
 RUN apt-get install -y g++-multilib
-RUN apt-get install -y clang-6.0
+RUN apt-get install -y clang-7
+RUN ln -s clang-7 /usr/bin/clang
```
sas: I don't think this is needed because I don't think we ever call `clang` directly.

Support/Testing/CircleCI/Dockerfile
```diff
@@ -65,7 +95,9 @@ RUN update-alternatives --set x86_64-w64-mingw32-gcc /usr/bin/x86_64-w64-mingw32
 RUN update-alternatives --set x86_64-w64-mingw32-g++ /usr/bin/x86_64-w64-mingw32-g++-posix
 
 # Install test dependencies
-RUN apt-get install -y git lldb-6.0 gdb clang-format-6.0 make dejagnu
+RUN apt-get install -y git lldb-7 gdb clang-format-7 make dejagnu
+RUN ln -s lldb-7 /usr/bin/lldb
```
sas: Ditto.

Support/Testing/CircleCI/Dockerfile
```diff
@@ -78,10 +110,14 @@ COPY Support/Scripts/common.sh /tmp
 COPY Support/Scripts/prepare-android-ndk.py /tmp
 RUN apt-get install -y unzip
 RUN /tmp/prepare-android-ndk.py
+RUN ln -s /tmp/android-sdk-linux/platform-tools/adb /usr/bin/adb
 
-# Install a version of cmake that is at least the minimum version we support.
-COPY Support/Testing/CircleCI/install-cmake.sh /tmp
-RUN /tmp/install-cmake.sh
+# Install latest cmake
+ADD https://cmake.org/files/v3.12/cmake-3.12.1-Linux-x86_64.sh /cmake-3.12.1-Linux-x86_64.sh
```
sas: You should update the script and use it instead.

Support/Testing/CircleCI/Dockerfile
```diff
@@ -78,10 +110,14 @@ COPY Support/Scripts/common.sh /tmp
 COPY Support/Scripts/prepare-android-ndk.py /tmp
 RUN apt-get install -y unzip
 RUN /tmp/prepare-android-ndk.py
+RUN ln -s /tmp/android-sdk-linux/platform-tools/adb /usr/bin/adb
```
sas: Do we need this?

Support/Scripts/run-lldb-tests.sh
lanza: Meh, do you actually think it's worth maintaining which version of llvm broke a test? e.g. making a 70 and 80 file. Would value would that provide?

Support/Testing/CircleCI/Dockerfile
lanza: Yea but I want to remove `clang-6.0` and `lldb-6.0` from usages throughout the project instead of manually having to update them every llvm release.

Support/Scripts/run-lldb-tests.sh
lanza: Not a permanent change. Just need to fix the patches given that the files changed in lldb such that the patch no longer works.

PR #680, opened 2018-09-06
Add some convenience utilities for ds2 docker image testing

lanza:


Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,6 +40,35 @@ MAINTAINER Stephane Sezer <sas@fb.com>
 RUN apt-get update
 RUN apt-get install -y software-properties-common python-software-properties wget
 
+# Incase lldb needs built
```
sas: Incase -> In case

Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,6 +40,35 @@ MAINTAINER Stephane Sezer <sas@fb.com>
...
+RUN apt-get install -y libz-dev
+RUN apt-get install -y swig
+RUN apt-get install -y ncurses-dev
```
sas: You can put these on just one line, it'll be faster to install.

`RUN apt-get install -y libz-dev ncurses-dev swig`

Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,6 +40,35 @@ MAINTAINER Stephane Sezer <sas@fb.com>
...
+
+# Make the developer's life not suck
+RUN apt-get -y install sudo
```
sas: Same here, do the apt-add-repositry/apt-get update first, then do one big apt-get update for all the "utility" packages.

Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,6 +40,35 @@ MAINTAINER Stephane Sezer <sas@fb.com>
...
+RUN apt-get -y install curl
+RUN apt-get -y install vim
+RUN apt-add-repository -y ppa:neovim-ppa/stable
+RUN apt-get update -y
+RUN apt-get -y install neovim
+
+# Debugging tools
+RUN apt-get install -y strace
+RUN apt-get install -y htop
+RUN apt-get install -y psmisc
```
sas: `RUN apt-get install -y htop psmisc strace`

Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,6 +40,35 @@ MAINTAINER Stephane Sezer <sas@fb.com>
...
+
+# Python 3.7 and misc
+RUN add-apt-repository -y ppa:deadsnakes/ppa
+RUN apt-get update -y
+RUN apt-get install -y python3.7
+RUN curl https://bootstrap.pypa.io/get-pip.py | python3.7
```
sas: Is this required? Isn't there just a `pip` package in apt, maybe from the deadsnakes repo?

Support/Testing/CircleCI/Dockerfile
```diff
@@ -40,6 +40,35 @@ MAINTAINER Stephane Sezer <sas@fb.com>
...
+RUN python3.7 -m pip install --upgrade pip
+RUN apt-get install -y python-pip
+RUN python -m pip install --upgrade pip
```
sas: Why do we need to upgrade `pip` right after installing it?

Support/Testing/CircleCI/Dockerfile
```diff
@@ -52,6 +81,7 @@ RUN apt-get install -y flex bison
 # Install x86 compilers
 RUN apt-get install -y g++-multilib
 RUN apt-get install -y clang-6.0
+RUN ln -s clang-6.0 /usr/bin/clang
```
sas: Is `/usr/local/bin` in `$PATH` by default? If it is, it might be better to create these symlinks there to avoid potential conflicts in `/usr/bin` in the future. Also, generally speaking, `/usr/bin` is managed by the distribution's package manager and `/usr/local` by the user.

Support/Testing/CircleCI/Dockerfile
xiaobai: Why do we need sudo? You should have root within the image itself, no?

Support/Testing/CircleCI/Dockerfile
lanza: Grammar nazi

Support/Testing/CircleCI/Dockerfile
lanza: When I have to work on the container for prolonged periods of time I find it worth it to set up a user and import my `~/.dotfiles` settings. Things don't work properly without a non-root user. I'm tempted to add a `adduser facebook` for this purpose.

Support/Testing/CircleCI/Dockerfile
lanza: Apt installs pip9 and while the latest release is pip18.

Support/Testing/CircleCI/Dockerfile
lanza: That is true, I'll change the paths.

Support/Testing/CircleCI/Dockerfile
lanza: `apt install python3-pip` installs to `/usr/local/lib/python3.5/dist-packages`

Support/Testing/CircleCI/Dockerfile
lanza: For the time being, this isn't necessary a "testing Dockerfile" but rather a "testing and debugging Dockerfile". Maybe once we clear out the blacklists we can strip it towards being testing-only but for now minimizing this file serves literally 0 use or benefit at the cost of making debugging more cumbersome.

lanza:
My rationale for these changes: 
For the time being, this isn't necessary a "testing Dockerfile" but rather a "testing and debugging Dockerfile". Maybe once we clear out the blacklists we can strip it towards being testing-only but for now minimizing this file serves literally 0 use or benefit at the cost of making debugging more cumbersome.

sas:
> For the time being, this isn't necessary a "testing Dockerfile" but rather a "testing and debugging Dockerfile". Maybe once we clear out the blacklists we can strip it towards being testing-only but for now minimizing this file serves literally 0 use or benefit at the cost of making debugging more cumbersome.

Fair enough

lanza:
@sas is that an approve?

PR #681, opened 2018-09-06
Change DS2LOG to use the proper compiler builtin function string

lanza:
DS2LOG uses __FUNCTION__ which does not accurately refer to the
parent function when you are working with a lambda as shown from this
error message:

    ds2/Sources/Target/Common/ProcessBase.cpp:149:7: warning: inside a
    lambda, '__FUNCTION__' expands to the name of the function call
    operator; consider capturing the name of the enclosing function
    explicitly [misc-lambda-function-name]
          DS2LOG(Debug,
          ^

MSVC's __FUNCTION__ behaves properly while GCC & clang require
__PRETTY_FUNCTION__ to replicate this. Conditionally using the right
version allows us to remove __DS2_LOG_CLASS_NAME__ as well.

### DIFF
--- a/Headers/DebugServer2/Utils/Log.h
+++ b/Headers/DebugServer2/Utils/Log.h
@@ -50,14 +50,16 @@ void Log(int level, char const *classname, char const *funcname,
 
 #define PRI_PTR_CAST(VAL) ((uintptr_t)(VAL))
 
-#if defined(__DS2_LOG_CLASS_NAME__)
-#define DS2LOG(LVL, ...)                                                       \
-  ds2::Log(ds2::kLogLevel##LVL, __DS2_LOG_CLASS_NAME__, __FUNCTION__,          \
-           __VA_ARGS__)
+#if defined(COMPILER_MSVC)
+#define FUNCTION_NAME __FUNCTION__
+#elif defined(COMPILER_CLANG) || defined(COMPILER_GCC)
+#define FUNCTION_NAME __PRETTY_FUNCTION__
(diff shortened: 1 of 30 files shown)


Headers/DebugServer2/Utils/Log.h
```diff
@@ -50,15 +50,15 @@ void Log(int level, char const *classname, char const *funcname,
...
-#define DS2LOG(LVL, ...)                                                       \
-  ds2::Log(ds2::kLogLevel##LVL, __DS2_LOG_CLASS_NAME__, __FUNCTION__,          \
-           __VA_ARGS__)
-#else
-#define DS2LOG(LVL, ...)                                                       \
-  ds2::Log(ds2::kLogLevel##LVL, nullptr, __FUNCTION__, __VA_ARGS__)
+#if defined(COMPILER_MSVC)
+#define FUNCTION_NAME __FUNCTION__
+#elif defined(COMPILER_CLANG) || defined(COMPILER_GCC)
+#define FUNCTION_NAME __PRETTY_FUNCTION__
```
xiaobai: You probably want a final thing else here. Something like:
```
#else
#error "Unsupported Compiler"
#endif
```

sas: Good point. There are examples of that elsewhere in the code, Look up the message we use and use the same probably.

(Although, to be fair, even with no "else" case, the build would fail because of `FUNCTION_NAME` being undefined.)

PR #682, opened 2018-09-06
Add .clang-tidy and fix a few issues it detects

lanza:
clang-tidy detects a few minor issues. Fix them here.

### DIFF
--- a/Sources/Utils/POSIX/Stringify.cpp
+++ b/Sources/Utils/POSIX/Stringify.cpp
@@ -16,14 +16,12 @@
 
 #include <cerrno>
 #include <csignal>
-// clang-format off
-#include <sys/types.h>
-#include <sys/ptrace.h>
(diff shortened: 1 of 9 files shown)


Sources/Utils/POSIX/Stringify.cpp
```diff
@@ -17,8 +17,8 @@
 #include <cerrno>
 #include <csignal>
 // clang-format off
-#include <sys/types.h>
-#include <sys/ptrace.h>
+#include <sys/types.h> // NOLINT
```
sas: Why is this required?

lanza: I don't know the fundamental reason why, but you added the `clang-format off` in 2016 because types needed to be included before ptrace.

lanza: That worked.

PR #683, opened 2018-09-06
Add a script to run clang-tidy over the codebase

lanza:
The command assumes that there exists a compile_commands.json in the
current directory. If there isn't then you must pass the
`--ccdb path-to-db` flag

### DIFF
--- a/Support/Scripts/tidy.py
+++ b/Support/Scripts/tidy.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3.7
+
+import argparse
+import functools
+import json
+import multiprocessing
+import os
+import subprocess
+import sys
+from typing import Optional
+
+# Some files should not be ran through clang-tidy for various reasons e.g. they
+# are generated with flex/bison and thus aren't guaranteed to be tidy. Pass
+# the `--ignore arg` flag where arg is a substring present in the entire file
+# name within the ccdb. e.g. to ignore
+# `/Users/lanza/Projects/ds2/mbuild/JSObjects/tokenizer.c` you ccan use
+# `--ignore tokenizer`.
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--ignore",
+        action="append",
+        default=[],
+        help="files listed in the ccdb that clang-tidy should ignore",
+    )
+    args = parser.parse_args()
+    return args
+
+
+def run_clang_tidy(ignores: [str], file: str) -> Optional[subprocess.CompletedProcess]:
+    for ignore in ignores:
+        if ignore in file:
+            return None
+    return subprocess.run("clang-tidy -warnings-as-errors='*' " + file, shell=True)
+
+
+def get_parent_ccdb_from(path: str) -> Optional[str]:
+    current_directory = path
+    while current_directory is not "/":
+        if os.path.exists(os.path.join(current_directory, "compile_commands.json")):
+            return os.path.join(current_directory, "compile_commands.json")
+        else:
+            current_directory = os.path.dirname(current_directory)
+    return None
+
+
+def die():
+    print(
+        """
+clang-tidy requires that a compile_commands.json is located in a parent
+directory of the file to be checked. If you build in `build` then you should
+run `ln -s build/compile_commands.json .`"""
+    )
+    sys.exit(1)
+
+
+def main() -> int:
+    args = parse_args()
+
+    ccdb_path = get_parent_ccdb_from(os.path.abspath(os.curdir))
+    if ccdb_path is None:
+        die()
+
+    with open(ccdb_path) as ccdb:
+        compile_commands = json.load(ccdb)
+
+        # test to see if there is a ccdb in a parent directory of the first
+        # source file since all files should have a common ccdb in a common
+        # ancestor
+        if get_parent_ccdb_from(compile_commands[0]["file"]) is None:
+            die()
+
+        files_to_parse = map(lambda x: x["file"], compile_commands)
+
+        run_clang_tidy_with_ignores = functools.partial(run_clang_tidy, args.ignore)
+
+        cpus = multiprocessing.cpu_count()
+        with multiprocessing.Pool(cpus) as p:
+            results = p.map(run_clang_tidy_with_ignores, files_to_parse)
+            filtered_results = filter(lambda x: x is not None, results)
+            return_codes = map(lambda x: x.returncode, filtered_results)
+            return any(return_codes)
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


lanza:
So running this detects errors in the parser and linter generated by flex and bison for JSONObjects so I can't add a circle-ci test along with clang-format.

Support/Scripts/tidy.py
```diff
@@ -0,0 +1,41 @@
...
+        for entry in compile_commands:
+            result = subprocess.run(["clang-tidy", entry["file"]], check=True)
+            if result is False:
+                passed = False
+
+        return 0 if passed else 1
+
+
+if __name__ == "__main__":
+    if not main():
```
sas: `sys.exit(main())` is cleaner.

Support/Scripts/tidy.py
```diff
@@ -0,0 +1,41 @@
...
+
+# This script assumes that there is a compile_commands.json in the current
+# directory. If there is not, you must pass 
+# `--ccdb path/to/compile_commands.json`.  
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--ccdb",
+        default="./compile_commands.json",
```
sas: I think if you remove the `./` you keep the same behavior but make your code more portable (Windows...).

Support/Scripts/tidy.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    args = parser.parse_args()
+    return args
+
+
+def main() -> int:
+    args = parse_args()
+    with open(args.ccdb) as ccdb:
+        compile_commands = json.load(ccdb)
+        os.chdir(compile_commands[0]["directory"])
```
sas: This assumes the working directory is the same for every command, which isn't necessarily true. You should move the `os.chdir()` inside the loop. If performance becomes an issue you can implement a check that verifies the directory neesd to be changed before actually changing it, but I'm pretty sure that won't be required.

Support/Scripts/tidy.py
```diff
@@ -0,0 +1,41 @@
...
+
+        passed = True
+        for entry in compile_commands:
+            result = subprocess.run(["clang-tidy", entry["file"]], check=True)
+            if result is False:
```
sas: `passed = passed and result`

sas:
Just saw your comment about not being able to enable this in CircleCI. Can we just have an exclude list, maybe as command-line arguments?

Support/Scripts/tidy.py
lanza: It is unless we add subprojects. All files in our ccdb are compiled out of the build directory e.g. `ds2/build`. 

Also, the purpose of the cd here is that `clang-tidy` reads out of `.`. So you need to be in a directory with a ccdb, not necessarily in the directory that the build actually happens in since cmake generates the ccdb with absolute paths to source files and dependencies and only relative paths for the output.

lanza:
Okay I added an `ignore` argument that can be passed `tokenizer` and `parser`.

lanza:
I'll add this to CircleCI later. Right now it would conflict with too many other changes that would require me to rebase over and over.

Support/Scripts/tidy.py
sas: yeah but we have subprojects in ds2 already.

If you need to `chdir()` in the directory where the ccdb is, why not do `os.chdir(os.path.dirname(args.ccd))`? It seems better to do explicitely "move to the directory where ccdb is" instead of having implicit behavior like this. Or am I missing something?

Support/Scripts/tidy.py
```diff
@@ -0,0 +1,68 @@
...
+
+# This script assumes that there is a compile_commands.json in the current
+# directory. If there is not, you must pass
+# `--ccdb path/to/compile_commands.json`.
+
+# Some files should not be ran through clang-tidy for various reasons e.g. they
+# are generated with flex/bison and thus aren't guaranteed to be tidy. Pass
+# the `--ignore arg` flag where arg is a substring present in the entire file
+# name within the ccdb. e.g. to ignore
+# `/Users/lanza/Projects/ds2/mbuild/JSObjects/tokenizer.c` you ccan use
```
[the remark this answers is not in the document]
lanza: The ccdb stores a complete path. "/Users/sas/source/ds2/the-right-build-subdirectory/Tools/JSOBjects/tokenizer.c" is unreasonable to have to type for this.

Support/Scripts/tidy.py
```diff
@@ -0,0 +1,68 @@
...
+def main() -> int:
+    args = parse_args()
+    with open(args.ccdb) as ccdb:
+        compile_commands = json.load(ccdb)
+        os.chdir(compile_commands[0]["directory"])
+
+        files_to_parse = list(map(lambda x: x["file"], compile_commands))
+        files_to_remove = []
+
+        for ignored_file in args.ignore:
```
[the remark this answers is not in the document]
lanza: True. I have a better idea.

Support/Scripts/tidy.py
lanza: > yeah but we have subprojects in ds2 already.

Oh yea, that is true. So I was wrong when I said

> It is unless we add subprojects. All files in our ccdb are compiled out of the build directory e.g. ds2/build.

It always uses `ds2/build` as value the "directory" key in the ccdb. And this is confirmed by looking at a ccdb from llvm. So using the value for the directory key in any entry in a ccdb is valid.

That being said, an explicit path via `os.chdir(os.path.abspath(os.path.dirname(args.ccdb)))` is cleaner.

Support/Scripts/tidy.py
```diff
@@ -0,0 +1,68 @@
...
+
+        for file in files_to_remove:
+            files_to_parse.remove(file)
+
+        all_runs_passed = True
+        cpus = multiprocessing.cpu_count()
+        with multiprocessing.Pool(cpus) as p:
+            results = p.map(run_clang_tidy, files_to_parse)
+            return_codes = map(lambda x: x.returncode, results)
+            return any(return_codes)
```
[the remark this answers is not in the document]
sas: Yeah you're relying on the implicit cast from boolean to int, I see.

Support/Scripts/tidy.py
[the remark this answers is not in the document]
sas: I think the way this kind of stuff is usually handled is by using something like `os.path.abspath()` or `os.path.realpath()`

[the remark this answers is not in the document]
lanza: I meant for manual usage. While the absolute path would be more idealistic for a scripted environment if you ever wanted to use this manually it's much nicer to use `--ignore tokenizer` as I have been doing while writing this.

PR #694, opened 2018-10-28
Remove erroneous arch setters in cmakelists

lanza:
For some reason there was duplicate code to set the arch in the main
CMakeLists.txt for Windows which was also incorrect as it didn't handle
Ninja at all or any combination of MSVC generators.

Remove the code as the arch is already set.


lanza:
Ah okay, I see the issue. cmake changed from using 

    cmake -G "Visual Studio 15 2017 Win64"

to specify an x86 compiler and a x64 build to

    cmake -G "Visual Studio 15 2017" -A x64 -Thost=x64 

where the `-A` and `-Thost` flags do the obvious thing. 

[See here](
https://cmake.org/cmake/help/v3.13/generator/Visual%20Studio%2015%202017.html#platform-selection).

lanza:
Okay this works now @sas

Support/CMake/Toolchain-Windows-x86_64-Toolset-x86_64.cmake
```diff
@@ -0,0 +1,11 @@
+##
```
sas: It seems to me this file is unused. You should get rid of it.

Generally speaking, in this project, we don't really care about which toolset is being used, so not specifying it at all and letting cmake pick the default seems like a better solution.

.appveyor.yml
```diff
@@ -9,16 +9,14 @@ branches:
 
 environment:
   matrix:
-    - generator: "Visual Studio 14 2015"
-      toolchain: "Windows"
     - generator: "Visual Studio 15 2017"
```
sas: The generator doesn't need to be in a variable and repeated anymore given that it's always the same. You should just get rid of this variable and put the generator name directly in the cmake command below.

Support/CMake/Toolchain-WinStore-ARM-Toolset-x86.cmake
```diff
@@ -0,0 +1,27 @@
...
+## Copyright (c) 2014-present, Facebook, Inc.
+## All rights reserved.
+##
+## This source code is licensed under the University of Illinois/NCSA Open
+## Source License found in the LICENSE file in the root directory of this
+## source tree. An additional grant of patent rights can be found in the
+## PATENTS file in the same directory.
+##
+
+# This is meant to be used on Windows, to generate Visual Studio project files.
```
sas: This comment is both outdated and repeated many times.

Support/CMake/Toolchain-WinStore-ARM-Toolset-x86_64.cmake
```diff
@@ -0,0 +1,28 @@
+##
```
sas: This file is completely unnecessary, let's ignore the toolset BS and just let cmake pick the default.

Support/CMake/Toolchain-Windows-x86_64-Toolset-x86_64.cmake
lanza: Dead.

.appveyor.yml
lanza: Done.

Support/CMake/Toolchain-WinStore-ARM.cmake
```diff
@@ -8,18 +8,12 @@
...
-# will build for WinStore ARM.
 set(CMAKE_SYSTEM_NAME WindowsStore)
 set(CMAKE_SYSTEM_VERSION 10.0)
 
 # For WinStore we build as a library because creating a process seems
 # restricted, so we just have a separate process open `ds2.dll` and run `main`
 # from it.
 set(LIBRARY 1)
+
+set(CMAKE_GENERATOR_PLATFORM ARM)
```
sas: Put this with line 11 and 12. Group the info that belongs together.

Support/CMake/Toolchain-WinStore-x86.cmake
```diff
@@ -0,0 +1,17 @@
...
+## All rights reserved.
+##
+## This source code is licensed under the University of Illinois/NCSA Open
+## Source License found in the LICENSE file in the root directory of this
+## source tree. An additional grant of patent rights can be found in the
+## PATENTS file in the same directory.
+##
+
+set(CMAKE_SYSTEM_NAME WindowsStore)
+set(CMAKE_SYSTEM_VERSION 10.0)
```
sas: You might wanna set all the variables explicitly in all toolchain files instead of relying on implicit value. That's just a personal preference thing though, so feel free to ignore.

PR #696, opened 2018-10-28
Add cmake install for ds2

lanza:
ds2 is a routinely used binary on Windows that's not attached to our internal toolchain. Having an install target is useful.


lanza:
I built and then copy and pasted it onto my path a few times before remembering that cmake had official support for this.

PR #697, opened 2018-11-02
Rename windows headers for -Wnonportable-system-include-path

lanza:
Change headers.h to Headers.h
Change psapi.h to Psapi.h
Change shlwapi to Shlwapi
Change ws2tcpip to WS2tcpip

### DIFF
--- a/Headers/DebugServer2/Base.h
+++ b/Headers/DebugServer2/Base.h
@@ -12,7 +12,7 @@
 
 #if defined(_WIN32)
 // clang-format off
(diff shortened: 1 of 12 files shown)


xiaobai:
I personally don't care about MinGW, so I'll let @sas make that call.

sas:
@mingw is there because it allows us to get gcc-specific warnings in Windows-specific code. I'm not too sure if I care about renaming these headers with a different case given that the filesystem on Windows is not case sensitive.

PR #698, opened 2018-11-02
Add advapi32 if linking with lld-link

lanza:
For some reason I didn't bother to look into, linking with lld-link
requires this library is explicitly added via target_link_libraries.


xiaobai:
What happens if you don't add this? I understand it "won't work" but what are the error/warning messages you get?

lanza:
> What happens if you don't add this? I understand it "won't work" but what are the error/warning messages you get?

Five or six unresolved symbols. `__imp__DoSomethingWindowsy can not be resolved`.

PR #701, opened 2018-11-05
Use Windows compilation flags when compiler id is Clang and MSVC

lanza:
Using clang-cl would cause the CMakeLists.txt to use the clang/gcc
flags which don't map properly to windows flags (e.g. /Wall on MSVC
is equal to -Weverything on clang).

Switch to clang-cl using Windows compilation flags via adding a check
for MSVC in the if statement.


xiaobai:
This approach seems fine to me. You can also do this with `CMAKE_CXX_SIMULATE_ID`. If you need finer grain control later on, you can check for `CMAKE_CXX_COMPILER_ID` matching "clang" and `CMAKE_CXX_SIMULATE_ID` matching "MSVC" to know for certain that you are dealing with clang-cl. See https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_SIMULATE_ID.html

lanza:
Very fancy. 👍🏼

CMakeLists.txt
```diff
@@ -394,7 +394,8 @@ if (TIZEN)
 endif ()
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR
-    "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+    "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" AND
+    NOT "${CMAKE_CXX_SIMULATE_ID}" MATCHES MSVC)
```
sas: Also, quote your strings.
```suggestion
    NOT "${CMAKE_CXX_SIMULATE_ID}" MATCHES "MSVC")
```

Tools/JSObjects/CMakeLists.txt
```diff
@@ -43,7 +43,9 @@ set_property(TARGET jsobjects PROPERTY C_STANDARD 99)
 set_property(TARGET jsobjects PROPERTY C_EXTENSIONS OFF)
 set_property(TARGET jsobjects PROPERTY CXX_STANDARD 11)
 set_property(TARGET jsobjects PROPERTY CXX_EXTENSIONS OFF)
-if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" OR
+    CMAKE_CXX_COMPILER_ID MATCHES "Clang" 
+    AND NOT CMAKE_CXX_SIMULATE_ID MATCHES MSVC)
```
sas: Here too.

lanza:
I didn't even know you could do parens in cmake if statements. It looks so wrong lol

PR #702, opened 2018-11-07
Add a WinStore-x86_64 cmake toolchain

lanza:


lanza:
We build iit on sandcastle

PR #705, opened 2018-12-20
Add pedantic and fix numerous pedantic errors

lanza:
lldb uses -Wpedantic and thus incorporating ds2 builds into lldb
requires building with pedantry

### DIFF
--- a/Sources/Core/ARM/HardwareBreakpointManager.cpp
+++ b/Sources/Core/ARM/HardwareBreakpointManager.cpp
@@ -19,17 +19,17 @@ namespace ds2 {
 
 int HardwareBreakpointManager::hit(Target::Thread *thread, Site &site) {
   return -1;
-};
+}
 
 ErrorCode HardwareBreakpointManager::enableLocation(Site const &site, int idx,
                                                     Target::Thread *thread) {
   return kErrorUnsupported;
-};
+}
 
 ErrorCode HardwareBreakpointManager::disableLocation(int idx,
                                                      Target::Thread *thread) {
   return kErrorUnsupported;
-};
+}
 
 size_t HardwareBreakpointManager::maxWatchpoints() {
   return _process->getMaxWatchpoints();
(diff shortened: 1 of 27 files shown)


Sources/Utils/Stringify.cpp
```diff
@@ -12,6 +12,9 @@
 #include "DebugServer2/Utils/Stringify.h"
 #include "DebugServer2/Utils/Log.h"
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcovered-switch-default"
```
xiaobai: This breaks the project. :(

Sources/Core/ARM/HardwareBreakpointManager.cpp
```diff
@@ -19,17 +19,17 @@ namespace ds2 {
 
 int HardwareBreakpointManager::hit(Target::Thread *thread, Site &site) {
   return -1;
-};
```
sas: The changes in this file are good. I'm not sure wtf these semicolons are doing there...

sas: The changes in this file are good. I'm not sure wtf these semicolons are doing there...

Sources/GDBRemote/DebugSessionImpl.cpp
```diff
@@ -206,9 +206,6 @@ ErrorCode DebugSessionImplBase::queryStopInfo(Session &session, Thread *thread,
   case StopInfo::kEventKill:
     DS2ASSERT(stop.reason == StopInfo::kReasonNone);
     break;
-
```
sas: This is fine too, I guess.

Sources/GDBRemote/DummySessionDelegateImpl.cpp
```diff
@@ -31,7 +31,7 @@ namespace GDBRemote {
 DummySessionDelegateImpl::DummySessionDelegateImpl() = default;
 
 size_t DummySessionDelegateImpl::getGPRSize() const {
-  DS2ASSERT(!"this method shall be implemented by inheriting class");
```
sas: Instead of using `false`, you should make this a `DS2BUG()`.

PR #706, opened 2019-07-17
Identity is not the same things as equality in Python

cclauss:
Use ==/!= to compare str, bytes, and int literals.
```
>>> slashes = "/"
>>> slashes += "/"
>>> slashes == "//"
True
>>> slashes is "//"
False
```
[flake8](http://flake8.pycqa.org) testing of https://github.com/facebook/ds2 on Python 3.7.1

$ __flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__
```
./Support/Scripts/tidy.py:41:11: F632 use ==/!= to compare str, bytes, and int literals
    while current_directory is not "/":
          ^
1     F632 use ==/!= to compare str, bytes, and int literals
1
```


sas:
Thanks for the change.

